PR #97, opened 2017-11-22
try to auto detect branch for queue build command

### LINKED ISSUE #92: 'vsts build queue' should queue for current branch if in a Git repository
`vsts build queue` would be much more useful if it defaulted to the current branch if the user is in a git repository. Also, having a `git` alias for it would make sense.

bergmeister:
to resolve #92 
Similar to the way how it is done for pull requests
Note that I am not sure if I could compile it correctly locally.


tedchamb:
@bergmeister Thanks for the contribution!  Your change is now in the dev branch.

Issue #88, opened 2017-11-20
Branch parameter inconsistent


bergmeister:
`vsts build queue` has a `--source-branch` parameter but `vsts list` has a `--branch` parameter.
I prefer to have the shorter `--branch` parameter but what counts at the end is consistency

tedchamb:
Agree this should be --branch.

PR #99, opened 2017-11-27
change name of --source-branch arg to just --branch on build queue. For PR create set target branch to the default branch regardless of whether 'detect' is on.

tedchamb:
…For PR create set target branch to the default branch regardless of whether 'detect' is on.


bergmeister:
Thanks. Is this sufficient to close #88 ?

PR #181, opened 2018-08-03
Use knack.log for logging, to fix incorrect logging levels (warnings not showing up)

zarenner:


zarenner:
Note that anything logged under a namespace besides the 'cli' namespace (https://github.com/Microsoft/knack/blob/master/knack/log.py#L13) will still have different/incorrect logging settings and probably won't show up unless --debug is specified. I'm not sure the best way to fix this but I don't think it's a big deal since there isn't much emitted in other namespaces at levels higher than debug.

PR #186, opened 2018-08-06
Now that logger.exception correctly emits to CLI, don't do it for CLIError

zarenner:
Without this PR, user errors are being emitted twice - first as a full callstack due to this logger.exception, and then later in Knack as an error.

### DIFF
--- a/src/common_modules/vsts-cli-common/vsts/cli/common/_credentials.py
+++ b/src/common_modules/vsts-cli-common/vsts/cli/common/_credentials.py
@@ -55,7 +55,6 @@ def clear_credential(team_instance):
     try:
         keyring.delete_password(key, _USERNAME)
     except keyring.errors.PasswordDeleteError as ex:
-        logger.exception(ex)
         raise CLIError('The credential was not found')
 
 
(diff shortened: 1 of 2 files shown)


src/common_modules/vsts-cli-common/vsts/cli/common/_credentials.py
```diff
@@ -55,7 +55,6 @@ def clear_credential(team_instance):
     try:
         keyring.delete_password(key, _USERNAME)
     except keyring.errors.PasswordDeleteError as ex:
-        logger.exception(ex)
```
zarenner: I removed this one since it's almost certainly user error. There were other logger.exception calls in common code that I didn't remove since it wasn't obvious if they were user errors or not. There are also logger.exception calls in specific product areas that I didn't touch (but can if you'd like).

src/common_modules/vsts-cli-common/vsts/cli/common/exception_handling.py
```diff
@@ -12,9 +12,9 @@
 logger = get_logger(__name__)
 
 def handle_command_exception(exception):
```
zarenner: This small change will cause only non user errors to have their callstack emitted.
By defining our own exception_handler on VstsCLI, I could go a step further and nuke the ugly boilerplate try/except blocks on each command that call this.

Knack already has logic for emitting CLIError as an error rather than exception: https://github.com/Microsoft/knack/blob/master/knack/cli.py#L159

PR #187, opened 2018-08-07
Don't use JSONDecodeError, not available on Python 2.7

zarenner:


zarenner:
Do we have any tooling set up to check for python2 and 3 compat?

PR #191, opened 2018-08-07
Switch from logger.exception to logger.debug to log exceptions in --d…

zarenner:
…ebug only

Move exception handling logic up to VstsCLI


src/vsts-cli/vsts/cli/vsts_cli.py
```diff
@@ -83,6 +87,21 @@ def get_runtime_version(self):  # pylint: disable=no-self-use
...
+        if isinstance(ex, VstsClientRequestError):
+            ex = CLIError(ex)
+
+        # Modify auth errors to be CLIError and have a helpful message
+        if isinstance(ex, VstsAuthenticationError):
+            ex = get_authentication_error(ex)
+
+        # Knack doesn't emit stacktraces for CLIErrors, but we want them on debug
+        if isinstance(ex, CLIError):
+            logger.debug(ex, exc_info=True)
```
zarenner: (knack does logger.error for CLIError, but we still want the stack trace on debug)

PR #193, opened 2018-08-08
fix python2 bug, missing arguments in super call: TypeError: super() takes at least 1 argument

tedchamb:


zarenner:
Thanks for the catch - I think there might be another of these or two hiding in our packaging code. I'll take a look.

PR #199, opened 2018-08-17
Universal Packages: Fix logging output when ArtifactTool output unparsable, by always emitting stderr as string not byte array

zarenner:
#### Actual
```
Universal Packages is currently in preview.
Failed to parse structured output from Universal Packages tooling (ArtifactTool)
Exception: the JSON object must be str, not 'bytes'
Log line: b'Failed to load x\xcf\x8d\x01, error: libunwind.so.8: cannot open shared object file: No such file or directory\n'
```

#### Expected
```
Universal Packages is currently in preview.
Failed to parse structured output from Universal Packages tooling (ArtifactTool)
Exception: Expecting value: line 1 column 1 (char 0)
Log line: b'Failed to load x/\x97\x01, error: libunwind.so.8: cannot open shared object file: No such file or directory\n'


zarenner:
Closing for now as I don't think this is quite the right fix. Will open a new PR shortly.

PR #219, opened 2018-10-12
Update install.py to include --user pip argument

rorpage:
I was unable to install this locally unless I added the `--user` pip argument.

Thanks!


geverghe:
Hi @rorpage - thank you for your contribution. We have created the [Azure DevOps Extension for Azure CLI](https://github.com/Microsoft/azure-devops-cli-extension). VSTS CLI has been deprecated and we will not be adding any more new features to it. 

For the Azure DevOps extension, you would need to install Azure CLI and then add the Azure DevOps extension. Check out the [Quick Start instructions](https://github.com/Microsoft/azure-devops-cli-extension#quick-start). I am closing this PR as this is no longer applicable to the Azure DevOps extension.

PR #240, opened 2018-11-14
Support python3 in dev_setup.py

vbossica:
Similar to how `azure-cli` works, support Python3 when installing `vsts-cli` locally.


vbossica:
Closing this PR as it is not needed with the new azure-cli extension.

PR #242, opened 2018-11-15
Install script to include SHA256 checksum verification for curl install

atbagga:


atbagga:
@willsmythe This was an old change where checksum verification was commented out. Any idea if this was due to some failure on some platform or just to avoid failures on missing to update the script on every change in python install script.

scripts/curl_install/README.md
```diff
@@ -3,17 +3,24 @@ Curl Install Script Information
 
 The scripts in this directory are used for installing through curl and they point to the packages on PyPI.
```
gauravsaralMs: assuming that the change is tested

PR #245, opened 2018-11-20
update to new storage account

gauravsaralMs:


atbagga:
There is another reference to https://vstscli.blob.core.windows.net in packaged_releases/windows/scripts/build_local.cmd.

gauravsaralMs:
> There is another reference to https://vstscli.blob.core.windows.net in packaged_releases/windows/scripts/build_local.cmd.

can't find. the other one is vstscli.azurewebsites... 
we are evaluating if that also needs to move or not

PR #259, opened 2018-11-29
Removing keyrings.alt and fixing urls for rebranding change

ishitam8:
Since, keyrings.alt package is  less secure, removing dependency on keyrings.alt package and storing credentials in file when keyring is missing.
This is done to support some linux distributions where keyring is not available.
A personalAccessTokens file will be created at path $HOME/.vsts/cli , which will be used to store the PAT tokens

Also fixing service url to include organization name for new azure devops URLs


atbagga:
Add some description to the PR, like why we are doing this.

PR #262, opened 2018-12-03
PS script to upload pypi packages

atbagga:


snankani:
would it make sense to have a mode for the script which can print the list of potential wheels it may publish ? This can help the user to get an understanding of the wheels he may end up publishing once the script runs.

gauravsaralMs:
in vsts cli repo, most of the scripts are py scripts.. 
ps1 seems to be odd one out 
not that I have any issues with that :) but just pointing out

gauravsaralMs:
I am not very sure if this kind of script should go in repo or in the CI/CD pipeline somewhere
because most of the contributors/ developers can't/ won't run it..

like normal contributes cannot update the pypi for vsts packages, it will always be just a handful of users who can do that

PR #281, opened 2018-12-17
update setup metadata

gauravsaralMs:


azure-devops-extension/setup.py
```diff
@@ -54,9 +54,9 @@
     license='MIT',
     description="Azure DevOps Command Line Interface",
     long_description=README + '\n\n' + HISTORY,
-    author="Microsoft Corporation",
-    author_email="vstscli@microsoft.com",
-    url="https://github.com/Microsoft/vsts-cli",
+    author="Microsoft",
+    author_email="VSTS_Social@microsoft.com",
+    url="https://github.com/Microsoft/vsts-cli/tree/azuredevopscli-dev",
```
atbagga: Why to change url? we plan to make this default branch anyways and then the repo url will point to the new code.

gauravsaralMs: till we make this default branch

PR #288, opened 2018-12-19
Setting default org on login and clear on logout

atbagga:


gauravsaralMs:
not very sure if login should alter default setting (just looks odd)
may be we should just show a message (in login flow) telling that "if" you want to set this as your default Org then run this command..
OR
we should have a bool parameter in login asking if user wants to set new instance as default instance

and
[very very corner case] login + logout flow will not restore state to start so if user does not remember the old default instance then it is gone

atbagga:
> not very sure if login should alter default setting (just looks odd)
> may be we should just show a message (in login flow) telling that "if" you want to set this as your default Org then run this command..

Doesn't sound like a good solution for people using only one organization. See the reported issue.. multiple people expected this behaviour already.

> OR
> we should have a bool parameter in login asking if user wants to set new instance as default instance
> and
> [very very corner case] login + logout flow will not restore state to start so if user does not remember the old default instance then it is gone

Login + logout will restore to start state. We are clearing the org in logout if it is the same as default.

@geverghe Thoughts?

Issue #217, opened 2018-10-11
Artifacttool doesn't work on docker image 


rorpage:
The latest version on Docker Hub is `0.1.3`: https://hub.docker.com/r/microsoft/vsts-cli/tags/

Here's the page I'm referring to: https://docs.microsoft.com/en-us/cli/vsts/install?view=vsts-cli-latest

atbagga:
@tedchamb Is there a reason we specify a specific docker version tag, rather than 'latest'?

rorpage:
If I may chime in, I pulled 0.1.3 as the Universal Packages piece of the CLI wasn't in 0.1.0.

I can open a separate issue for this, but the `artifacttool` doesn't appear to install correctly in 0.1.3. Perhaps it's .NET Core and the base is Python Alpine?

tedchamb:
@atbagga I agree, we shouldn't be specifying the version in those commands.  @wnjenkin, is this something you can update?

@zarenner, can you investigate the `artifacttool` issue?  Let me know if we need to create a new image.

rorpage:
@tedchamb @wnjenkin Is this something I could do for Hacktoberfest? I'm also more than happy to help @zarenner with a new image too!

rorpage:
Here's the message I get in my container:

```
Universal Packages is currently in preview.
[Errno 2] No such file or directory: '/root/.vsts/cli/tools/artifacttool/ArtifactTool_linux-x64_0.2.74/artifacttool': '/root/.vsts/cli/tools/artifacttool/ArtifactTool_linux-x64_0.2.74/artifacttool'
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/knack/cli.py", line 197, in invoke
    cmd_result = self.invocation.execute(args)
  File "/usr/local/lib/python3.6/site-packages/knack/invocation.py", line 172, in execute
    cmd_result = parsed_args.func(params)
  File "/usr/local/lib/python3.6/site-packages/knack/commands.py", line 105, in __call__
    return self.handler(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/knack/commands.py", line 203, in _command_handler
    result = op(client, **command_args) if client else op(**command_args)
  File "/usr/local/lib/python3.6/site-packages/vsts/cli/package/common/universal.py", line 41, in publish_package
    return artifact_tool.publish_universal(team_instance, feed, name, version, description, path)
[...]

PR #290, opened 2018-12-21
Send Linux distro name and version when resolving ArtifactTool to download

zarenner:
Part of fix for #217

Basically we need to send the distro name up to blobstore service, so that it can send down the alpine-x64 variant of ArtifactTool rather than the linux-x64 variant.

Tested on Alpine, makes request:
`vsts.vss_client : GET https://vsblob.dev.azure.com/mseng/_apis/clienttools/ArtifactTool/release?distroVersion=3.8.1&distroName=alpine&arch=x86_64&osName=Linux`

### DIFF
--- a/src/common_modules/vsts-cli-package-common/setup.py
+++ b/src/common_modules/vsts-cli-package-common/setup.py
@@ -17,6 +17,7 @@ VERSION = '0.1.4'
 
 REQUIRES = [
     'colorama>=0.3.9',
+    'distro==1.3.0',
     'humanfriendly==4.7',
     'knack==0.4.1',
     'python-dateutil==2.7.3',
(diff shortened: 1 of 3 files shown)


zarenner:
Thanks, I'll target that branch as well. For my future knowledge, is dev effectively dead or just temporarily forked? Also, I didn't internally see vsts-cli listed as a governed repository, so please let me know if you'll need anything to move the OSS approval forward.

atbagga:
Dev is effectively dead - mostly yes. We are not planning on releasing another version unless really necessary. 
OSS approval will follow up on email with you.

zarenner:
I'm going to abandon this one, since I have no plans to support alpine for deprecated vsts-cli.
Opened #324 for azure devops cli extension.

PR #296, opened 2018-12-27
add commands to manage Git refs

vbossica:
This changeset adds methods to manage Git refs:
* list of references
* create/update/delete a single reference


azure-devops/azext_devops/dev/repos/commands.py
```diff
@@ -47,7 +53,8 @@ def load_code_commands(self, _):
         g.command('create', 'create_pull_request', table_transformer=transform_pull_request_table_output)
         g.command('update', 'update_pull_request', table_transformer=transform_pull_request_table_output)
         g.command('show', 'show_pull_request', table_transformer=transform_pull_request_table_output)
-        g.command('list', 'list_pull_requests', table_transformer=transform_pull_requests_table_output)
+        g.command('list2', 'list_pull_requests', table_transformer=transform_pull_requests_table_output)
```
gauravsaralMs: is this intentional ?
cc: @atbagga

azure-devops/azext_devops/test/repos/test_ref.py
```diff
@@ -0,0 +1,74 @@
...
+    def tearDown(self):
+        self.mock_list_refs.stop()
+        self.mock_update_ref.stop()
+
+    def test_list_refs(self):
+        response = list_refs(devops_organization=self._TEST_DEVOPS_ORGANIZATION,
+                             project='sample project',
+                             detect='off')
+        # assert
+        self.mock_list_refs.assert_called_once()
```
atbagga: We are trying to move to more strict tests. See #318. 

In your case the commands in itself don't have any code logic just call the sdk. You can use **assert_called_once_with()**  to make more strict asserts. 

Anything else that you think would help or any suggestions in this area are welcome.

azure-devops/azext_devops/dev/repos/ref.py
```diff
@@ -0,0 +1,110 @@
...
+                                  old_object_id=object_id)
+        return client.update_refs([ref_update], repository, project)[0]
+    except VstsServiceError as ex:
+        raise CLIError(ex)
+
+
+def update_ref(name, old_object_id, new_object_id, repository=None, devops_organization=None,
+               project=None, detect=None):
+    """Update a reference.
+    :param str name: Name of the reference to create (example: heads/my_branch).
```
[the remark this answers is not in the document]
ishitam8: Same for delete_ref command as well

azure-devops/azext_devops/dev/repos/ref.py
```diff
@@ -0,0 +1,110 @@
...
+                               filter=filter)
+    except VstsServiceError as ex:
+        raise CLIError(ex)
+
+
+def create_ref(name, object_id, locked=False, repository=None, devops_organization=None, project=None, detect=None):
+    """Create a reference.
+    :param str name: Name of the reference to create (example: heads/my_branch or tags/my_tag).
+    :param str object_id: id of the object to create the reference from.
+    :param bool locked: If the reference is locked (default False)
```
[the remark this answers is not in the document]
vbossica: You're right. I think the problem is in the serialization/deserialization of boolean values (which is strange if accurate.)
When retrieving the list of all refs, the `isLocked` value is `null`:
```
"isLocked": null,
"isLockedBy": null,
"name": "refs/tags/tag_4",
"objectId": "4f20eea9a98c52b75e8264634ca63b662d6fa55a",
```
when a boolean value should be true/false.

Suggestion is to remove the `--locked` parameter to unblock this PR and open an issue for further investigation.

azure-devops/azext_devops/dev/repos/_format.py
```diff
@@ -252,6 +252,8 @@ def _transform_ref_row(row):
             table_row['Old Object ID'] = old_id
             table_row['New Object ID'] = new_id
     table_row['Name'] = get_ref_name_from_ref(row['name'])
+    table_row['success'] = row['success']
```
atbagga: nit: Success
nit: Update Status 

for user visible strings in the table header.

azure-devops/azext_devops/dev/repos/arguments.py
```diff
@@ -57,3 +57,7 @@ def load_code_arguments(self, _):
     with self.argument_context('repos import create') as context:
         context.argument('repository', options_list=('--repository', '-r'))
         context.argument('git_source_url', options_list=('--git-source-url', '--git-url'))
+
+    with self.argument_context('repos ref') as context:
+        context.argument('repository', options_list=('--repository', '-r'))
```
atbagga: nit: Defining 'repository' alias is not needed now. We have added it to the parent command group so it applies to all commands now.

azure-devops/azext_devops/test/repos/test_ref.py
```diff
@@ -0,0 +1,82 @@
...
+    # Attempt to load mock (works on Python version below 3.3)
+    from mock import patch, ANY
+
+from azext_devops.dev.common.services import clear_connection_cache
+from azext_devops.dev.repos.ref import (list_refs, create_ref, delete_ref, update_ref)
+
+
+class TestRefMethods(unittest.TestCase):
+
+    _TEST_DEVOPS_ORGANIZATION = 'https://AzureDevOpsCliTest.visualstudio.com'
```
atbagga: Put an invalid url here like.. https://some-invalid-org.visualstudio.com. 

This will fail the test since the "client = get_git_client(organization)" call in each of the commands makes a network call. You will need to mock it, in your case to return a GitClient object. 

We discovered this recently and I am fixing other tests. 

This one is implemented correctly. 
https://github.com/Microsoft/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/test/pipelines/test_pipelines_build.py

check -  self.get_client = patch('vsts.vss_connection.VssConnection.get_client')

tests/recordings/test_ref_createListDelete.yaml
```diff
@@ -0,0 +1,124 @@
+interactions:
```
atbagga: The yaml is collected on a test which was run without the cache cleared. This will fail in recording if the test is run on a system with cache cleared.

azure-devops/azext_devops/dev/repos/ref.py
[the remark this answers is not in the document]
atbagga: Go ahead with the suggestion.

azure-devops/azext_devops/dev/repos/arguments.py
vbossica: It's the `repos pr` group that has the `repository` alias, not the `repos` one.

azure-devops/azext_devops/dev/repos/ref.py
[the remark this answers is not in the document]
vbossica: the issue lies somewhere between the `vsts-python-api` and the rest endpoint: https://github.com/Microsoft/azure-devops-python-api/issues/174

azure-devops/azext_devops/dev/repos/ref.py
```diff
@@ -0,0 +1,131 @@
...
+        organization, project, repository = resolve_instance_project_and_repo(
+            detect=detect,
+            organization=organization,
+            project=project,
+            repo=repository)
+        client = get_git_client(organization)
+        # by default, the create method does not support setting the is_locked value
+        # to True.
+        ref_update = GitRefUpdate(is_locked=False,
+                                  name=resolve_git_refs(name),
```
gauravsaralMs: Why are we not using [resolve_git_ref_heads](https://github.com/Microsoft/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/dev/common/git.py#L115)

atbagga:
Thank you @vbossica for your contribution! 👍

PR #306, opened 2019-01-03
Telemetry related fixes

atbagga:
1. Add support for logging shell type in CI (code from azure cli)
1. Fix issue in vss_Connection cache - case mismatch was resulting in cache miss
1. Telemetry is getting logged in UTs (needs to be fixed - get_vss_connection call should not be made in UTs), this PR handles not to log the telemetry
1. Added tests

### DIFF
--- a/azure-devops/azext_devops/dev/common/telemetry.py
+++ b/azure-devops/azext_devops/dev/common/telemetry.py
@@ -88,7 +88,7 @@ def _send_tracking_ci_event(devops_organization=None, ci_client=None):
         logger.debug(ex, exc_info=True)
         return False
 
-
+# azure cli uses this to get shell type from os environment
 def _get_shell_type():
     import os
     if 'ZSH_VERSION' in os.environ:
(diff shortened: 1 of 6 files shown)


azure-devops/azext_devops/dev/common/telemetry.py
```diff
@@ -87,4 +89,30 @@ def _send_tracking_ci_event(devops_organization=None, ci_client=None):
...
+    import os
+    if 'ZSH_VERSION' in os.environ:
+        return 'zsh'
+    if 'BASH_VERSION' in os.environ:
+        return 'bash'
+    if 'KSH_VERSION' in os.environ or 'FCEDIT' in os.environ:
+        return 'ksh'
+    if 'WINDIR' in os.environ:
+        return 'cmd'
+    return _remove_cmd_chars(_remove_symbols(os.environ.get('SHELL')))
```
gauravsaralMs: not able to get this..
can you put some comments here, for future readers as well

PR #308, opened 2019-01-03
Fixing error message for login with invalid token

ishitam8:
An organization with public projects enabled will not throw any exception for invalid token. Hence, handling anonymous user case here in login command.

### DIFF
--- a/azure-devops/azext_devops/dev/team/credentials.py
+++ b/azure-devops/azext_devops/dev/team/credentials.py
@@ -31,7 +31,9 @@ def credential_set(devops_organization=None):
         connection = _get_vss_connection(devops_organization, credentials)
         location_client = connection.get_client('vsts.location.v4_1.location_client.LocationClient')
         try:
-            location_client.get_connection_data()
+            connection_data = location_client.get_connection_data()
+            if connection_data.authenticated_user.id == _ANONYMOUS_USER_ID:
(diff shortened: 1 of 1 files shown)


azure-devops/azext_devops/dev/team/credentials.py
```diff
@@ -31,10 +31,15 @@ def credential_set(devops_organization=None):
         connection = _get_vss_connection(devops_organization, credentials)
         location_client = connection.get_client('vsts.location.v4_1.location_client.LocationClient')
         try:
-            location_client.get_connection_data()
+            connection_data = location_client.get_connection_data()
         except Exception as ex2:
             logger.debug(ex2, exc_info=True)
             raise CLIError("Failed to authenticate using the supplied token.")
+        else:
```
gauravsaralMs: This `else` is for which `if` ?

gauravsaralMs: I still think we can remove this because `except` anyways returns control using `raise` but nothing v. important

ishitam8: That's for try block. It would be executed if there are no exceptions raised in try block.

atbagga: +1 for removing else. 
indentation should be sufficient to achieve it.

PR #324, opened 2019-01-09
ArtifactTool: Send distro when determining artifactool to download

zarenner:


zarenner:
Part of #217

azure-devops/azext_devops/dev/artifacts/artifacttool_updater.py
```diff
@@ -12,6 +12,7 @@
 import uuid
 import zipfile
 
+import distro
```
gauravsaralMs: where is this package getting installed (so that DevOps extension can use it) ?

azure-devops/setup.py
```diff
@@ -18,6 +18,7 @@
 # http://pypi.python.org/pypi/setuptools
 
 REQUIRES = [
+    'distro==1.3.0',
```
gauravsaralMs: This package has a very scoped usage in the code, instead of bringing/ installing it with overall extension can we wait for it's first time usage before installing it?
we are evaluating something similar at 
 https://github.com/Microsoft/vsts-cli/pull/323 
do share your thoughts

PR #325, opened 2019-01-10
Adding Team Related Commands

gauravsaralMs:


azure-devops/azext_devops/dev/team/_format.py
```diff
@@ -59,9 +59,54 @@ def _transform_service_endpoint_row(row):
...
+
+
+def transform_team_table_output(result):
+    table_output = [_transform_team_row(result)]
+    return table_output
+
+
+def _transform_team_row(row):
+    table_row = OrderedDict()
+    table_row['ID'] = row['id']
```
snankani: I think we should have constant variables for these.

azure-devops/azext_devops/dev/team/team.py
```diff
@@ -0,0 +1,155 @@
...
+                                                                    devops_organization=devops_organization,
+                                                                    project=project)
+        core_client = get_core_client(devops_organization)
+        team_to_create = WebApiTeam(name=name, description=description)
+        return core_client.create_team(team=team_to_create, project_id=project)
+    except VstsServiceError as ex:
+        raise CLIError(ex)
+
+
+def delete_team(team_id, devops_organization=None, project=None, detect=None):
```
snankani: return type (rtype) missing in docstring for delete_team, get_team, get_teams and get_team_members methods

azure-devops/azext_devops/test/team/test_team.py
```diff
@@ -0,0 +1,126 @@
...
+    # Attempt to load mock (works on Python version below 3.3)
+    from mock import patch
+
+from vsts.core.v4_0.core_client import CoreClient
+from azext_devops.dev.team.team import (create_team,
+                                        delete_team,
+                                        get_team,
+                                        get_teams,
+                                        get_team_members,
+                                        update_team)
```
snankani: Given that all the functions from team.py are imported why not use * here??

from azext_devops.dev.team.team import *

azure-devops/azext_devops/test/team/test_team.py
```diff
@@ -0,0 +1,126 @@
...
+        self.mock_get_client.return_value = CoreClient(base_url=self._TEST_DEVOPS_ORGANIZATION)
+
+        #clear connection cache before running each test
+        clear_connection_cache()
+
+    def tearDown(self):
+        patch.stopall()
+
+    def test_create_team(self):
+        create_team(self._TEST_TEAM_NAME, self._TEST_TEAM_DESCRIPTION, self._TEST_DEVOPS_ORGANIZATION, self._TEST_PROJECT_NAME, 'Off')
```
snankani: 'off' can be declared in a variable

azure-devops/azext_devops/test/team/test_team.py
```diff
@@ -0,0 +1,126 @@
...
+
+        #assert
+        self.mock_get_team_members.assert_called_once()
+        get_team_members_param = self.mock_get_team_members.call_args_list[0][1]
+        self.assertEqual(self._TEST_TEAM_NAME, get_team_members_param['team_id'], str(get_team_members_param))
+        self.assertEqual(self._TEST_PROJECT_NAME, get_team_members_param['project_id'], str(get_team_members_param))
+        self.assertEqual(10, get_team_members_param['top'], str(get_team_members_param))
+        self.assertEqual(2, get_team_members_param['skip'], str(get_team_members_param))
+
+    def test_update_team(self):
```
snankani: Can we add a test case for update_team where both name and description are none?

azure-devops/azext_devops/dev/team/team.py
gauravsaralMs: done, delete_team will continue to not have any return type

azure-devops/azext_devops/test/team/test_team.py
gauravsaralMs: done

azure-devops/azext_devops/dev/team/team.py
```diff
@@ -0,0 +1,155 @@
...
+    :type description: str
+    :param devops_organization: Azure Devops organization URL. Example: https://dev.azure.com/MyOrganizationName/
+    :type devops_organization: str
+    :param project: Name or ID of the project.
+    :type project: str
+    :param detect: When 'On' unsupplied arg values will be detected from the current working
+                   directory's repo.
+    :rtype: :class:`<WebApiTeam> <core.v4_0.models.WebApiTeam>`
+    """
+    if name is None and description is None:
```
gauravsaralMs: don't see much gain in doing this.
Reasons:
1. This same check is done by API
2. Code in CLI is not operating on these and is simply passing the same to service
3. This should be a corner case (without data here, but this is not the happy path) so for every valid command this check is happening twice (which will slow down every valid command)

I would have recommended to add this check if
we were doing some operation on these variables like splitting, length or anything 
but right now this does not seem useful.

azure-devops/azext_devops/dev/team/_format.py
gauravsaralMs: Opened 
https://github.com/Microsoft/vsts-cli/issues/330 to track this

azure-devops/azext_devops/dev/team/team.py
gauravsaralMs: i am fine either ways just that it looks like a performance optimization for negative scenario (which might be a good thing) but `if name is None and description is None:` will execute always (for positive scenario) as well which might not be a good thing

gauravsaralMs: ok. now I see the point of this check..
it is actually to tell user about requirement rather than fail fast, user can't do mapping from API error to command line

ishitam8: Yeah, the exception message from API  won't be enough for the user to know about the missing arguments

azure-devops/azext_devops/dev/team/_format.py
atbagga: Suggestion: Shouldn't they be defined in the same file? Since they are coming from json which is specific to command outputs in this file. Otherwise as u said tracking casing and all will be an issue.

azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -28,3 +28,9 @@ def load_team_help():
     short-summary: Manage service endpoints/service connections
     long-summary: https://docs.microsoft.com/en-us/rest/api/azure/devops/serviceendpoint/endpoints
     """
+
+    helps['devops team'] = """
+    type: group
+    short-summary: Manage teams
+    long-summary: https://docs.microsoft.com/en-us/rest/api/azure/devops/core/teams
```
atbagga: Shouldn't there be atleast some text with the link? e.g. _For More info refer documentation here -_

azure-devops/azext_devops/dev/team/team.py
```diff
@@ -0,0 +1,158 @@
...
+    """Create a team.
+    :param name: Name of the new team.
+    :type name: str
+    :param description: Description of the new team.
+    :type description: str
+    :param devops_organization: Azure Devops organization URL. Example: https://dev.azure.com/MyOrganizationName/
+    :type devops_organization: str
+    :param project: Name or ID of the project.
+    :type project: str
+    :param detect: When 'On' unsupplied arg values will be detected from the current working
```
atbagga: (Default: on)
*at other places too.

azure-devops/azext_devops/dev/team/team.py
```diff
@@ -0,0 +1,158 @@
...
+        core_client = get_core_client(devops_organization)
+        team_to_create = WebApiTeam(name=name, description=description)
+        return core_client.create_team(team=team_to_create, project_id=project)
+    except VstsServiceError as ex:
+        raise CLIError(ex)
+
+
+def delete_team(team_id, devops_organization=None, project=None, detect=None):
+    """Deletes a team.
+    :param team_id: The name or id of the team to delete.
```
atbagga: nit: should it be called _--team_ or _--team-id_, _--id_ since it accepts both name and id?
For project and repo we have like that.

azure-devops/azext_devops/dev/team/team.py
```diff
@@ -0,0 +1,158 @@
...
+    :param detect: When 'On' unsupplied arg values will be detected from the current working
+                   directory's repo.
+    :rtype: [WebApiTeam]
+    """
+    try:
+        devops_organization, project = resolve_instance_and_project(detect=detect,
+                                                                    devops_organization=devops_organization,
+                                                                    project=project)
+        core_client = get_core_client(devops_organization)
+        return core_client.get_teams(top=top, skip=skip, project_id=project)
```
atbagga: We had that long discussion on validating inputs like top, skip > 0. Are we planning to follow it?

azure-devops/azext_devops/test/team/test_team.py
```diff
@@ -0,0 +1,122 @@
...
+
+try:
+    # Attempt to load mock (works on Python 3.3 and above)
+    from unittest.mock import patch
+except ImportError:
+    # Attempt to load mock (works on Python version below 3.3)
+    from mock import patch
+
+from vsts.core.v4_0.core_client import CoreClient
+from azext_devops.dev.team.team import *
```
atbagga: From [pep8 guidelines](https://www.python.org/dev/peps/pep-0008/#imports) - _Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools._

azure-devops/azext_devops/test/team/test_team.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def test_create_team(self):
+        create_team(self._TEST_TEAM_NAME, self._TEST_TEAM_DESCRIPTION, self._TEST_DEVOPS_ORGANIZATION, self._TEST_PROJECT_NAME, self._OFF)
+        
+        #assert
+        self.mock_create_team.assert_called_once()
+        create_team_param = self.mock_create_team.call_args_list[0][1]
+        self.assertEqual(self._TEST_PROJECT_NAME, create_team_param['project_id'], str(create_team_param))
+        self.assertEqual(self._TEST_TEAM_NAME, create_team_param['team'].name, str(create_team_param))
+        self.assertEqual(self._TEST_TEAM_DESCRIPTION, create_team_param['team'].description, str(create_team_param))
```
atbagga: Is this any different from assert_called_once_with(param1, param2, param3)?

azure-devops/azext_devops/dev/team/_help.py
ishitam8: I agree with @atbagga . Also if we are adding the link to documentation, then it should be consistent across other commands as well. We are not adding documentation reference in any other commands except for '"service-endpoint" and "team" group.

azure-devops/azext_devops/dev/team/team.py
ishitam8: Yeah makes sense. My only concern is then commands would look a little odd (with team word twice in command instead of just id) , which is not the case at other places. Since project or repo name/id is passed as an optional or auto-detect argument in those commands.

For example : `az devops team show --team "XYZ Team" ` or
`az devops team delete --team "XYZ Team" `

@gauravsaralMs  , @geverghe  , @snankani . Thoughts?

azure-devops/azext_devops/dev/team/team.py
snankani: To be honest, I would prefer to have very basic input validations which are not api specific and having below validations in CLI looks convincing to me.
(1) trimming a string and validating if its length is greater than ZERO
(2) value greater than zero for parameters like top or skip

azure-devops/azext_devops/test/team/test_team.py
snankani: I remember leaving the comment earlier to use wildcard as all the functions of the class were being imported. Looking at the doc shared by Atul, apparently wildcard is not recommended, so let's move back to using explicit imports instead of wildcard usage.

ishitam8: Done.

azure-devops/azext_devops/test/team/test_team.py
ishitam8: Added a test with name and description both none.

azure-devops/azext_devops/dev/team/team.py
geverghe: I would suggest name and id; isn't name id always scoped to the command object that is being referred to? Is there any other universal name and ID? The only two universal concepts seem to be project and organization which are both referred to as --project --organization. Therefore, my suggestion would az devops team --name "XYZ Team" / az devops team --id 123

ishitam8: @geverghe  Are you suggesting to split the argument ? 
Actually a single argument would suffice here. Either name or id. Because API accepts either of them. There is no universal name and ID.  What Atul referred is done in artifacts [for feed argument](https://github.com/Microsoft/vsts-cli/blob/45242f0d300a94549ed5c58ebf2458bf4fdb0ba6/azure-devops/azext_devops/dev/artifacts/universal.py#L22)

azure-devops/azext_devops/dev/team/team.py
ishitam8: Yeah, makes sense to have basic validations across most commands. Opened #331 for the same.

tests/devopsTeamTest.py
```diff
@@ -0,0 +1,92 @@
...
+            assert create_team_output2["description"] == team_description
+
+            #list team command
+            list_teams_command = 'az devops team list --output json --detect off'
+            list_teams_output = self.cmd(list_teams_command).get_output_in_json()
+            assert len(list_teams_output) == 3  # one default project team and 2 teams created by test
+            verified_team_list = False
+            for team in list_teams_output:
+                if (team["id"] == created_team_id):
+                    verified_team_list = True
```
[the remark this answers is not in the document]
ishitam8: Updated the checks

azure-devops/azext_devops/dev/team/team.py
ishitam8: Updated the help text

PR #328, opened 2019-01-11
use run pip from azure cli core instead of having our own implementat…

gauravsaralMs:
use run pip from azure cli core instead of having our own implementation.

also added UT to make sure dependecy is checked in cli core so that we can react (as this is not a public method)

### DIFF
--- a/azure-devops/azext_devops/dev/common/pip_helper.py
+++ b/azure-devops/azext_devops/dev/common/pip_helper.py
@@ -3,9 +3,9 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-import pip
-
 from knack.log import get_logger
+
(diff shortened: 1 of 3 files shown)


azure-devops/azext_devops/dev/common/pip_helper.py
```diff
@@ -16,7 +16,5 @@ def install_keyring():
 
 def _install_package(package_name):
     logger.debug('installing %s', package_name)
-    if hasattr(pip, 'main'):
-        pip.main(['install', package_name])  # pylint: disable=no-member
-    else:
-        pip._internal.main(['install', package_name])  # pylint: disable=protected-access
+    pip_args = ['install', package_name]
+    _run_pip(pip_args)  # pylint: disable=protected-access
```
ishitam8: _run_pip returns 'status code' . We can throw some error if it fails.

PR #332, opened 2019-01-16
UT to test if keyring installation works or not

gauravsaralMs:


azure-devops/azext_devops/test/common/test_pip_helper.py
```diff
@@ -12,7 +13,26 @@ def test_run_pip_is_present_in_cli_core(self):
...
+        install_keyring()
+        # uninstall keyring
+        pip_args = ['uninstall', 'keyring', '-y']
+        from azure.cli.core.extension.operations import _run_pip
+        pip_status_code = _run_pip(pip_args)  # pylint: disable=protected-access
+        if pip_status_code > 0:
+            self.fail('keyring uninstall failed')
 
+        try:
+            from imp import reload
```
atbagga: https://stackoverflow.com/questions/961162/reloading-module-giving-nameerror-name-reload-is-not-defined

This suggested that reload is not in the same import for different python versions, but the tests passed for this PR in different env. Please check it.

The test is passing due to different exceptions, do a selecting catch not generic.

azure-devops/azext_devops/test/common/test_pip_helper.py
```diff
@@ -12,7 +13,26 @@ def test_run_pip_is_present_in_cli_core(self):
...
+            reload(keyring)
+            self.fail('importing keyring should have failed')
+        except:
+            print('expected')
+
+        # install keyring and make sure we can import that
+        install_keyring()
+        import keyring
```
atbagga: Shouldn't  this be reload?

gauravsaralMs:
not sure if there is a good way to test this, once a package is loaded it is impossible to unload it so closing this PR

PR #336, opened 2019-01-16
Adding YAML (and supporting scripts) so that we run pipelines using YAML

gauravsaralMs:


azure-pipelines-create-release.yml
```diff
@@ -0,0 +1,56 @@
...
+  pool:
+    vmImage: 'vs2017-win2016'
+
+  steps:
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: '3.7.0'
+      architecture: 'x64'
+
+  - script: python -m pip install --upgrade pip && pip install -r requirements.txt
```
ishitam8: requirements.txt has been removed. So, you can skip this step.

gauravsaralMs: done

PR #340, opened 2019-01-17
Fix usage of Id to make it consistent for commands in 'az devops **'

gauravsaralMs:
covering the commands under devops in phase 1

### DIFF
--- a/azure-devops/azext_devops/dev/team/arguments.py
+++ b/azure-devops/azext_devops/dev/team/arguments.py
@@ -44,8 +44,6 @@ def load_team_arguments(self, _):
     with self.argument_context('devops service-endpoint') as context:
         context.argument('devops_organization', options_list=('--organization', '--org'))
         context.argument('detect', **enum_choice_list(_ON_OFF_SWITCH_VALUES))
-    with self.argument_context('devops service-endpoint get') as context:
-        context.argument('service_endpoint_id', options_list='--id')
     with self.argument_context('devops service-endpoint create') as context:
         context.argument('service_endpoint_type', **enum_choice_list(_SERVICE_ENDPOINT_TYPE))
         context.argument('authorization_scheme', **enum_choice_list(_SERVICE_ENDPOINT_AUTHORIZATION_SCHEME))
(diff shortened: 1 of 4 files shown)


azure-devops/azext_devops/dev/team/project.py
```diff
@@ -94,40 +94,36 @@ def create_project(name, devops_organization=None, process=None, source_control=
...
             raise CLIError('Project deletion was cancelled.')
-        print('Deleted project {}'.format(project_id))
+        print('Deleted project {}'.format(id))
         return operation
     except VstsServiceError as ex:
         raise CLIError(ex)
 
 
-def show_project(project_id=None, name=None, devops_organization=None, detect=None, open_browser=False):
+def show_project(id=None, devops_organization=None, detect=None, open_browser=False):
```
atbagga: Id is compulsory now. Also now there is no check in the code to handle if it is none.

gauravsaralMs: nice catch made it madatory

azure-devops/azext_devops/dev/team/arguments.py
```diff
@@ -60,4 +57,3 @@ def load_team_arguments(self, _):
         context.argument('devops_organization', options_list=('--organization', '--org'))
         context.argument('detect', **enum_choice_list(_ON_OFF_SWITCH_VALUES))
         context.argument('project', options_list=('--project', '-p'))
-        context.argument('team_id', options_list=('--team-id', '--id'), type=str)
```
ishitam8: I think just adding '--id' here would be enough.
`context.argument('team_id', options_list=('--id'), type=str)`

This way we don't have to change the signature of the commands where just team_id is used. 
Also, it works for user because team_id, project _id is abstracted. And it would allign with API naming while writing the commands.

ishitam8: Also pylint won't throw any error in this case.

gauravsaralMs:
> What about "devops admin banner" commands? Since it falls under same group and not a separate pivot.

Will take it up separately, doing folder by folder

azure-devops/azext_devops/dev/team/arguments.py
ishitam8: As discussed offline, current changes makes more sense. Also reduces the overhead when adding a new command. Hence resolving this

PR #344, opened 2019-01-17
fail in case some style check fails

gauravsaralMs:


scripts/runStyleCheck.ps1
```diff
@@ -11,9 +13,22 @@ pip install flake8
 
 Write-Output "Running pylint on all source"
 pylint --rcfile pylintrc ./azure-devops/azext_devops -f colorized
+
+if ($LastExitCode -ne 0) {
+    $failure = $true
+  }
+
 Write-Output "Pylint OK!"
```
atbagga: Write-Output "Pylint OK!"

This should be in the else part. Same for flake8.

PR #347, opened 2019-01-18
use test_ so that pytest can discover recorded test files as test files

gauravsaralMs:
This will ensure 
1. little bit perf improvement
2. consolidates test result view
3. also running just pytest locally will make sure everything is picked up in single go


gauravsaralMs:
test_reposPrCommandsReviewersTest is behaving weirdly when ran with other tests so I have to keep it out for now, if someone has any pointers do let me know
still would like to consolidate things so that we don't have more tests like `test_reposPrCommandsReviewersTest` flowing into the system

PR #350, opened 2019-01-18
Changing behavior of Logout command

ishitam8:
Earlier behavior:
 1. az devops logout
 [Removes default token stored from credential store]
2. az devops logout --org https://dev.azure.com/<MY_ORGANIZATION>
[Remove credentials for MY_ORGANIZATION ]

Updated behavior:
 1. az devops logout
 **[Removes credentials of all organizations ]**
2. az devops logout --org https://dev.azure.com/<MY_ORGANIZATION>
[Remove credentials for MY_ORGANIZATION ]

Organization URLs are stored in file located in azuredevops config directory at user HOME path
(HOME\.\azure\azuredevops\organization_list)


azure-devops/azext_devops/dev/common/_credentials.py
```diff
@@ -29,18 +32,46 @@ def set_credential(devops_organization, token):
...
     cred_store.set_password(key, token)
 
+    if os.path.isfile(_ORGANIZATION_LIST_FILE):
+        # No need to add organization if it's already present.
+        with open(_ORGANIZATION_LIST_FILE, 'r') as org_list:
+            for organization in org_list:
+                if key == organization.rstrip():
+                    return
+    with open(_ORGANIZATION_LIST_FILE, 'a+') as org_list:
+        org_list.write(key + "\n")
```
atbagga: \n is platform specific. Some platforms have \r\n. 
I think if you use writeline method instead of write you wont need the \n ?

azure-devops/azext_devops/dev/common/_credentials.py
```diff
@@ -29,18 +32,46 @@ def set_credential(devops_organization, token):
...
+    
+    if key == _DEFAULT_CREDENTIAL_KEY:
+        # remove all organizations and delete the file
+        if os.path.isfile(_ORGANIZATION_LIST_FILE):
+            with open(_ORGANIZATION_LIST_FILE) as org_list_file:
+                for organization in org_list_file:
+                    cred_store.clear_password(organization.rstrip())
+            os.remove(_ORGANIZATION_LIST_FILE)
+        else:
+            raise CLIError('No credentials were found.')
```
atbagga: Breaking back compat: 
User will have stored creds but no file, after the upgrade user runs logout command. He will get no credentials were found and creds will not be cleared.

ishitam8: I agree, but what's the alternative here? We can't know what credentials were stored by user before the upgrade. The reason being keyring or credential manager doesn't provide a list of keys/credentials.

azure-devops/azext_devops/dev/team/credentials.py
```diff
@@ -51,7 +51,10 @@ def credential_clear(devops_organization=None):
     if devops_organization is not None:
         devops_organization = get_base_url(devops_organization)
     clear_credential(devops_organization)
-    print('The credential was successfully cleared.')
+    if devops_organization is not None:
+        print('The credential was successfully cleared.')
+    else:
+        print('Logged out of all azure devops organizations.')
     _check_and_clear_default_organization(devops_organization)
```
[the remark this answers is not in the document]
ishitam8: Nice catch, Thanks

azure-devops/azext_devops/dev/common/_credentials.py
ishitam8: Python actually takes care of translating new line to platform specific line separator. [Documentation here](https://docs.python.org/3/library/os.html#os.linesep)

@gauravsaralMs

azure-devops/azext_devops/dev/common/_credentials.py
atbagga: Agree we cannot delete all keys. 

Lets try to clear with default key atleast if it is present we will clean it.. if not, the message no credentials were found is also valid as per earlier behavior. 
We will end up covering what was supported earlier i.e. clearing the default key.

ishitam8: Yea, makes sense to at least clear the old default credential. Modified 'clear_credential' function. Thanks

PR #355, opened 2019-01-22
Parameter renames to fix auto generated documentation

atbagga:
devops_organization --> organization
open_browser --> open

### DIFF
--- a/azure-devops/azext_devops/dev/pipelines/build.py
+++ b/azure-devops/azext_devops/dev/pipelines/build.py
@@ -21,7 +21,7 @@ logger = get_logger(__name__)
 
 
 def build_queue(definition_id=None, definition_name=None, branch=None, variables=None, open_browser=False,
-                devops_organization=None, project=None, detect=None, source_branch=None, commit_id=None):
+                organization=None, project=None, detect=None, source_branch=None, commit_id=None):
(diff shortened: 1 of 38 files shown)


azure-devops/azext_devops/dev/pipelines/build.py
```diff
@@ -20,8 +20,8 @@
 logger = get_logger(__name__)
 
 
-def build_queue(definition_id=None, definition_name=None, branch=None, variables=None, open_browser=False,
```
gauravsaralMs: `open_browser` is much more verbose/ explicit.
I would recommend we retain `open_browser` as well and have `open` as one of the alias.
@geverghe  can tell better

geverghe: IMO, we should start with just --open;

gauravsaralMs: we can go ahead with just `open` but it is not very clear what is `open` in terms of CLI. `open_browser` was looking better to me that it was explicit that we are going to open something in browser.
but as long as we are relying on help text then it is fine

geverghe: Noted; we can probably start here and wait for feedback.

PR #359, opened 2019-01-22
Delete contributing.md

atbagga:
Please make sure the code is following contribution guidelines at doc/contributing.md


atbagga:
@geverghe There are two files with same name right now at root. Git did not create a conflict due to case change in the two files. 

Deleting the earlier one and updating the link in the readme to point to the new one.

atbagga:
> weird that this passed in our pipeline (it should fave failed in windows machine, if files differ only in casing)

I was expecting even local pull should have failed but seems like it just ignores one file.

Issue #364, opened 2019-01-22
Include YAML example for using Azure DevOps extension


geverghe:
Include an example on how Azure DevOps extension can be added and used using a YAML script.

PR #365, opened 2019-01-22
Users/geverghe/include yaml example

geverghe:
Please make sure the code is following contribution guidelines at doc/contributing.md

 - [x] : This PR has a corresponding issue open in the Repository. #364 
 - [ ] : Approach is signed off on the issue. - Documentation PR, hence NA

Included the example for using Azure DevOps using YAML configuration. Note that in the documentation the environment variable used is AZURE_DEVOPS_CLI_PAT. Need to update this post new release.

[Edit]: Was running into a few issues on Linux and windows with the initial scripts provided. Updated to verified working YAMLs. @atbagga , @gauravsaralMs, @ishitam8


doc/samples.md
```diff
@@ -5,6 +5,7 @@
 4. [Query ouput](samples.md#query-output)
 4. [Open items in browser](samples.md#open-items-in-browser)
 5. [Use the Azure DevOps Extension in a release pipeline](samples.md#use-the-azure-devops-extension-in-a-release-pipeline)
+6. [Use the Azure DevOps Extension with YAML]()
```
atbagga: Missing anchor link.

doc/samples.md
```diff
@@ -102,5 +103,60 @@ To use the Azure DevOps Extension in a hosted agent using a Release Pipeline, ex
 6. Add the script - either via file or inline. For the example, the script has been included inline. 
 ![PowerShell](/doc/images/Script.PNG)
 
-The first line of the script installs the Azure CLI and adds the Azure DevOps Extension. 
+Including the inline script for reference
+```bash
+pip install --pre azure-cli==2.0.55 --extra-index-url https://azurecliprod.blob.core.windows.net/edgeaz --h
+az --version
```
gauravsaralMs: we will need `az --version` only for debugging and while debugging we might need --debug in other commands as well

doc/samples.md
```diff
@@ -102,5 +103,60 @@ To use the Azure DevOps Extension in a hosted agent using a Release Pipeline, ex
...
+steps:
+- script: az --version
+  displayName: 'Show Azure CLI version'
+
+- script: az extension add -n azure-devops
+  displayName: 'Install Azure DevOps Extension'
+
+- script: echo ${AZURE_DEVOPS_CLI_PAT} | az devops login
+  env:
+    AZURE_DEVOPS_CLI_PAT: $(System.AccessToken)
```
gauravsaralMs: one problem I see with this script is that it is targeting the same org/ project where it is running, can we add example on how to pass these variables when targeting other accounts? 
because I think that will be a more prominent scenario to start with

doc/samples.md
geverghe: Considering it's a first glance into set up using YAML, are they required? The aim is to provide simplistic view to set things up.

doc/samples.md
geverghe: Replace $(System.AccessToken) with the PAT? Won't that make the PAT available in plain text?

doc/samples.md
gauravsaralMs: nope, not required

doc/samples.md
gauravsaralMs: I meant we can put this in variable in release/ build/ pipeline and use that instead of using system accesstoken

geverghe: Isn't that more details to explain? We need to explain setting up variables and then using it. Further, I am assuming the 80% case of using this would be go to a project -> create a new pipeline or from GitHub -> add app and create a new pipeline. In both, the system tokens should work.

PR #383, opened 2019-01-27
Fix markdownlint issues

vbossica:
Simply fix the markdownlint warnings


CONTRIBUTING.md
```diff
@@ -4,39 +4,39 @@ Thank you for your interest in contributing to the Azure DevOps Extension!
...
 👍 - upvote
 👎 - downvote
 
-#### Submit New Feature request
+### Submit New Feature request
 
 Use our Feature Request [template](https://github.com/Microsoft/azure-devops-cli-extension/issues/new/choose) to create a new feature request.
 
-# Get started
+## Get started
```
atbagga: All the content in this section is now at one level above that it should be. Please lower the heading hierarchy for all subitems in this section like Pull request etiquette etc.

atbagga: Do double check that out in other files as well. 
Due to pushing each top heading some  subitems have come to same level as the section heading.

gauravsaralMs:
If possible can we also add a check in PR pipeline (YAML) to make sure this does not break again?
cc: @atbagga

CONTRIBUTING.md
vbossica: Double-checked

atbagga:
> If possible can we also add a check in PR pipeline (YAML) to make sure this does not break again?
> cc: @atbagga

Created a task - not blocking the PR for this. #387

PR #384, opened 2019-01-28
Create Workitems in bulk

atbagga:
This script creates work items in bulk. 
It can take two params a text file with work item titles in each newline. 
A template with params like organisation, project, work item type etc filled in which will be used for all Workitems created.

In interactive mode the script works without any params, user can set the params interactively and then keep creating new work items with titles being asked for as user input.

### DIFF
--- a/examples/createWorkitemsInBulk.ps1
+++ b/examples/createWorkitemsInBulk.ps1
@@ -1,0 +1,62 @@
+#---------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+param(
+[string]$workitem_titles,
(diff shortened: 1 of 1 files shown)


examples/createWorkitemsInBulk.ps1
```diff
@@ -0,0 +1,62 @@
...
+    Write-Host("Creating a task with assigned to - " + $assignedto)
+}
+
+if (!$workitem_titles)
+{
+    Write-Host("No file for workitem titles. Will run in interactive mode to create workitems.")
+    while(1) {
+        $title = Read-Host("Enter workitem title: ")
+        $command_string = "az boards work-item create --org $organization --project $project --title ""$title"" --type $workitem_type" # --assigned-to $assignedto #Doesnot seem to work
+        Invoke-Expression($command_string)
```
gauravsaralMs: user is supposed to come out using Ctrl+C?

PR #386, opened 2019-01-28
Setup live test run pipeline

atbagga:
This PR includes changes for -

1. Adding pipeline yaml for live test run
1. Script to clean up the recordings to prepare the machine for live test run
1. Modify runTest.ps1 script to allow running only UTs or only recordings. By default it still continues to run both.
1. UTs modified to not make the network call for fetching the client.
1. UTs modified to not assert against get_credential call since it will not be made if PAT token is present on the machine.
1. Editing VCR test helper to set auth from Env VAR first and then fall back to hardcoded PAT.

 - [x] : This PR has a corresponding issue open in the Repository. #382 


gauravsaralMs:
may be should also add support that if `AZURE_TEST_RUN_LIVE` is set then recording tests will run in live mode, this is what azure live tests do

.azure-pipelines/azure-pipelines-live-test-run.yml
```diff
@@ -0,0 +1,51 @@
...
+      versionSpec: '$(python.version)'
+
+  - task: PowerShell@2
+    displayName: 'Setup test machine'
+    inputs:
+      targetType: 'filePath'
+      filePath: 'scripts/setupCiMachine.ps1'
+
+  # Run a Python script.
+  - task: PythonScript@0
```
gauravsaralMs: should we put this step in runTests.ps1?
reason being that in case I want to run the tests in live mode locally (against my org)
then I will have to do multiple steps..
so putting this as part of runtests.ps1 is looking more reasonable

tests/utilities/helper.py
```diff
@@ -17,6 +17,20 @@ def get_random_name(length):
...
     return wrapper
+
+
+def set_authentication(test_function):
+    def wrapper(*args, **kwargs):
+        print("Setting auth for test run.")
+        pat_from_env = os.environ.get(PAT_ENV_VARIABLE_NAME)
+        if not pat_from_env:
+            print("PAT not found in environment variable. Falling back to hardcoded token.")
+            os.environ[PAT_ENV_VARIABLE_NAME] = DEVOPS_CLI_TEST_PAT_TOKEN
```
gauravsaralMs: do we need this?

tests/utilities/helper.py
```diff
@@ -8,7 +8,7 @@
 import string
 
 DEVOPS_CLI_TEST_ORGANIZATION = 'https://AzureDevOpsCliTest.visualstudio.com'
```
gauravsaralMs: every one developing will not have permission on this org, so I think we should make this also configurable (pass target org and pat token to runtests.ps1) so that everyone can run tests in live mode against their own org.
I know our tests are not their right now that it can run against any collection (because some of them expect some setup)
but we should try to make our tests org agnostic in long run

.azure-pipelines/azure-pipelines-live-test-run.yml
atbagga: I want to  make all scripts python. So I do not want to add stuff to the powershell script.
If you still feel otherwise we can do that.

tests/utilities/helper.py
atbagga: Actually No.. I was discussing this with Vladimir too regarding getting rid of the PAT in the file which can get accidently raised in the PR and be public. 

Should we set it to hardcoded PAT so that pipelines running recorded tests do not need to change else we will have to edit all pipelines to have a PAT env which can be any random value.

tests/utilities/helper.py
atbagga: This org is not forced on all the tests. External contributors can define the org in their test and run the test and get a recording. Some with the test repo access will only need to rerun the test and update the recording with our org. 

Not considering this comment for this PR. You can raise a issue to track that.

atbagga:
> may be should also add support that if `AZURE_TEST_RUN_LIVE` is set then recording tests will run in live mode, this is what azure live tests do

We can do this when we combine the test run script into one. The script should take parameters to decide what tests to run (UTs, Recordings, live run). Keeping this outside the scope of this PR.

PR #396, opened 2019-01-30
Setup live test run pipeline

atbagga:
This PR includes changes for -

1. Adding pipeline yaml for live test run
1. Script to clean up the recordings to prepare the machine for live test run
1. Modify runTest.ps1 script to allow running only UTs or only recordings. By default it still continues to run both.
1. UTs modified to not make the network call for fetching the client.
1. UTs modified to not assert against get_credential call since it will not be made if PAT token is present on the machine.
1. Editing VCR test helper to set auth from Env VAR first and then fall back to hardcoded PAT.

 - [x] : This PR has a corresponding issue open in the Repository. #382 

Sample Pipeline run : https://dev.azure.com/AzureDevOpsCliOrg/AzureDevOpsCli/_build?definitionId=44

I will target this to master branch once this PR goes in.


tests/utilities/helper.py
```diff
@@ -7,8 +7,8 @@
 import random
 import string
 
-DEVOPS_CLI_TEST_ORGANIZATION = 'https://AzureDevOpsCliTest.visualstudio.com'
-DEVOPS_CLI_TEST_PAT_TOKEN = 'vj3ep2pg3fo6vxsklkwvkiy23dkbyynmfpg4vb66xniwr23zylla'
+DEVOPS_CLI_TEST_ORGANIZATION = 'https://dev.azure.com/azuredevopsclitest'
```
gauravsaralMs: may be for some other day
we have this constant across tests
so say I record my test against https://dev.azure.com/gsaral
then it will fail in recordings? because URI will not match (my assumption)

atbagga: Hmm.. noticed it while recording for the policy test after merge.

PR #399, opened 2019-01-31
Allow recording tests to run on any organization

atbagga:

With this change we will allow a developer to record a test against any organization. The live tests will run against the test org so the developer needs to write tests independent of the existing environment while recording.

We will need to add an environment variable in the live test run pipeline. 
AZURE_DEVOPS_EXT_TEST_ORG : Https://dev.azure.com/azuredevopsclitest


tests/test_reposPoliciesTest.py
```diff
@@ -8,13 +8,15 @@
 
 from azure.cli.testsdk import ScenarioTest
 from azure_devtools.scenario_tests import AllowLargeResponse
-from .utilities.helper import DEVOPS_CLI_TEST_ORGANIZATION, disable_telemetry, set_authentication
+from .utilities.helper import disable_telemetry, set_authentication, get_test_org_from_env_variable
+
+DEVOPS_CLI_TEST_ORGANIZATION = get_test_org_from_env_variable() or 'Https://dev.azure.com/baggaatul24'
```
atbagga: Just to verify I have run this one against the non test organization.

gauravsaralMs: 👍

Issue #272, opened 2018-12-10
--assigned-to throwing "Identity Value 'X' for field 'Assigned to' is an unknown identity


Gander7:
`vsts work item create --title "Test Title" --type "Bug" --assigned-to "user@company.com"`
throws following error
`The identity value 'XXXXXXXXXXXXXXXXXXX@Live.com' for field 'Assigned To' is an unknown identity.`

I've tried different formats for --assigned-to
- `Windows Live ID\\user@company.com`
- `user@company.com`
- `Full Name`

Throws error regardless of type or title.
I've tried various users in case it was a specific user.

Any thoughts or workarounds? We are trying to migrate to devops and this is a breaking issue for us :(

atbagga:
I only see "The identity value 'user-alpha@Live.com' for field 'Assigned To' is an unknown identity." when user-alpha does not have valid project/organization membership.

I have tried all the three formats like.. 
1. user-name@organization-name.com
2. user-name
3. "FirstName LastName"

and they work for me for a valid user. 

Can you try creating a workitem in the UI and assign to the same user as you are trying from CLI to help us troubleshoot further.

snankani:
@Gander7 We are looking for your reply to help you better. Request to take some time and let us know if the above suggested steps work for you. 

FYI, we have migrated to Azure DevOps extension which is an extension to Azure CLI. Please find details here https://github.com/Microsoft/azure-devops-cli-extension/blob/master/README.md

ishitam8:
We were able to reproduce this issue. Hence, reopening this.

PR #403, opened 2019-01-31
Fix identity issue for non-microsoft accounts

ishitam8:

 - [ ] : This PR has a corresponding issue open in the Repository. #272 


azure-devops/azext_devops/dev/boards/work_item.py
```diff
@@ -70,7 +70,7 @@ def create_work_item(work_item_type, title, description=None, assigned_to=None,
             if assigned_to == '':
                 resolved_assigned_to = ''
             else:
-                resolved_assigned_to = _resolve_identity_as_unique_user_id(assigned_to, organization)
+                resolved_assigned_to = resolve_identity_as_display_name(assigned_to, organization)
```
atbagga: Two email Ids resolving to same display name can cause issue. WIT API accepts the email id and name itself lets prefer that approach rather than trying to resolve ourself.

Issue with using display name: 

Command to assign an item to person X,
PS D:\GithubRepos\Work\MS\azure-devops-cli-extension\azure-devops> az boards work-item create --title "title" --type bug --assigned-to "lastname.firstname@**something**.com"

Wit gets assigned to person Y-
ID    Type    Title    Assigned To                        State
----  ------  -------  ---------------------------------  -------
81    Bug     title    Atul Bagga <lastname.firstname@**somethingelse**.com>  New

ishitam8: Nice catch!

azure-devops/azext_devops/dev/common/identities.py
```diff
@@ -110,5 +110,11 @@ def get_display_name_from_identity(identity):
     return identity.provider_display_name
 
 
+def get_account_from_identity(identity):
```
atbagga: Should  we have some test for this. If the strings in the API change we will not catch it.

ishitam8: I'll take this up in some another PR. Probably would add tests for another functions as well.

PR #414, opened 2019-02-04
fix merge pipeline

gauravsaralMs:


gauravsaralMs:
**Why merging with single approval?**
merging this so that in case some other change breaks something then we can catch it.
without this change out merge pipeline is currently broken

Issue #352, opened 2019-01-21
Bump keyring version from 13.2.1 to 15.1.0 or latest


snankani:

ishitam8:
Keyring version issue would be resolved with next extension release [i.e 0.4.0]

PR #419, opened 2019-02-05
Bump up keyring version to 17.1.1

ishitam8:

 - [ ] : This PR has a corresponding issue open in the Repository. #352 

### DIFF
--- a/azure-devops/azext_devops/dev/common/pip_helper.py
+++ b/azure-devops/azext_devops/dev/common/pip_helper.py
@@ -12,7 +12,7 @@ logger = get_logger(__name__)
 
 
 def install_keyring():
-    _install_package('keyring~=13.2.1')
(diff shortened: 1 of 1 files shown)


gauravsaralMs:
> Some code we had in https://github.com/Microsoft/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/dev/common/credential_store.py
> was dependent on the version of keyring package. We had written the code to handle the changes for chaining keyrings (which were introduced in keyring version 15 or 16). But lets double check it once in next bug bash.
> 
> Code:
> 
> ```
> def _only_builtin(backend):
>             return (
>                 backend.__module__.startswith('keyring.backends.') and
>                 'chain' not in backend.__module__
>             )
> ```

may be we can write a test case to see which keyring is backend is being used/ initialized with in any platform.. 
right now I don't think we have much coverage

PR #427, opened 2019-02-07
minor change in pip helper and steps to validate that it works

gauravsaralMs:

Changes include:
Install keyring and other packages to extension folder.
This is needed because in some systems user will not have permission to install packages at machine level (like azure shell)

Do not pack test when creating whl because we should not be shipping tests to client machines

While running UTs let it discover modules in extension folder. The local files module will still get preference over modules in extension dir


.azure-pipelines/azure-pipelines-merge.yml
```diff
@@ -148,28 +148,37 @@ jobs:
...
+
+  - task: PythonScript@0
+    displayName: 'Find Extension Directory'
+    inputs:
+      scriptSource: 'inline'
+      script: |
+        from azure.cli.core.extension import get_extension_path
+        extensionPath = get_extension_path('azure-devops')
+        print("##vso[task.setvariable variable=extensionPath]"+extensionPath)
+      failOnStderr: true
```
atbagga: A lot of inline script code we are adding in yml is making the pipelines complex. 

1. Whatever we need in the pipeline will also be required by the developer so it makes sense to add these scripts in the code repo itself.
2. Since the pipelines are harder to debug, using locally tested scripts in the pipelines will keep the pipelines up-to-date and more robust.

gauravsaralMs:
https://github.com/Microsoft/azure-devops-cli-extension/issues/438

PR #433, opened 2019-02-11
pack compressed VSTS package with CLI code for better install performance

gauravsaralMs:


atbagga:
Why name it vstsCompressed? vsts is sufficient I guess.

gauravsaralMs:
> Why name it vstsCompressed? vsts is sufficient I guess.

Just to make someone does not copies the whole VSTS folder from python api repo without compressing

atbagga:
> > Why name it vstsCompressed? vsts is sufficient I guess.
> 
> Just to make sure someone does not copies the whole VSTS folder from python api repo without compressing

@gauravsaralMs @ishitam8  This doesn't follow the standard of all lowercase names. Should it be vsts_compressed ?
I would have been fine with vsts also I don't think someone will replace that with original vsts package files blindly.

gauravsaralMs:
> > > Why name it vstsCompressed? vsts is sufficient I guess.
> > 
> > 
> > Just to make sure someone does not copies the whole VSTS folder from python api repo without compressing
> 
> @gauravsaralMs @ishitam8 This doesn't follow the standard of all lowercase names. Should it be vsts_compressed ?
> I would have been fine with vsts also I don't think someone will replace that with original vsts package files blindly.

@atbagga 
nice catch, feel free to change/fix or open an issue but given that this is just a temp thing till we get new optimized DevOps python client package I am not very much in favor of putting time on this again.

PR #448, opened 2019-02-18
trim how we are using same help text over and over again

gauravsaralMs:

### DIFF
--- a/azure-devops/azext_devops/dev/repos/pull_request.py
+++ b/azure-devops/azext_devops/dev/repos/pull_request.py
@@ -36,10 +36,6 @@ def show_pull_request(id, open=False, organization=None, detect=None):  # pylint
     :type id: int
     :param open: Open the pull request in your web browser.
     :type open: bool
-    :param organization: Azure Devops organization URL. Example: https://dev.azure.com/MyOrganizationName/
-    :type organization: str
-    :param detect: Automatically detect organization. Default is "on".
-    :type detect: str
     :rtype: :class:`GitPullRequest <git.v4_0.models.GitPullRequest>`
     """
     try:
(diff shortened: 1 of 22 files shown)


azure-devops/azext_devops/dev/team/arguments.py
```diff
@@ -22,27 +21,24 @@
                                           SERVICE_ENDPOINT_AUTHORIZATION_SERVICE_PRINCIPAL]
 
 
+# pylint: disable=line-too-long
```
atbagga: This one can be fixed. Line length limit is to improve readability. 
We should not disable all pylint warnings to fix CI :)

azure-devops/azext_devops/dev/repos/pull_request.py
```diff
@@ -82,10 +78,6 @@ def list_pull_requests(repository=None, creator=None, include_links=False, revie
     :type skip: int
     :param top: Maximum number of pull requests to list.
     :type top: int
```
atbagga: Project help text instance here in list_pull_requests command. 

Probably a one off miss but difficult to find in the diff if any more are missed. Just double check once before merging.

atbagga:
There are couple of comments but approving since the changes do not need a review again.

azure-devops/azext_devops/dev/team/arguments.py
gauravsaralMs: agree, but it's hard to validate style changes locally right now

azure-devops/azext_devops/dev/repos/pull_request.py
gauravsaralMs: agree, lets clean/fix it as and when we find anything

PR #460, opened 2019-02-25
Users/midenn/pipeline uploaddownload

mitchdenny:

The following PR adds download, upload and list support for Pipeline Artifacts in the extension. In order to test this out you will need to export the following environment variable to force the artifact tool logic to download a newer version of artifacttool with upload/download support.

```export AZURE_DEVOPS_EXT_ARTIFACTTOOL_OVERRIDE_VERSION=0.2.99```

That requirement is only until we made this newer build of artifact tool the default version in all scale units (we'll do that before we merge this PR).

The list support is capable of listing all artifact types, but upload and download support is limited to Pipeline Artifacts at this point of time (the check for this is implemented in artifact tool). Given we will want the same logic to work on-premises we should work with the Pipelines team to lay support for FCS on top of this PR.

The proposed usage for this command set is mentioned here #81, however instead of ```get``` I opted for ```download``` to align with the YAML syntax that we will be using in Pipelines. I additionally implemented an ```upload``` command which will work provided the user has permissions to edit build information (useful for external build engine integration scenarios).

As part of this change I moved the artifact tool modules up to the common area since that logic will now be shared across the Pipelines and Artifacts command-sets.

### CODE REVIEW

**Comment by geverghe** [comment]
Hey @mitchdenny - would this allow download of artifacts from builds as well? az pipelines build artifact download ?

**Comment by mitchdenny** [comment]
> Hey @mitchdenny - would this allow download of artifacts from builds as well? az pipelines build artifact download ?

It would allow download of Pipeline Artifacts from builds, but it doesn't yet support file containers. Started a thread with the Pipelines team on what they want here.

**Comment by mitchdenny** [comment]
@atbagga and @geverghe this requires another review. We've had some more discussions with the Pipelines teams and we will be adding the ability to download file containers using the runs artifacts download command.

However my plan is to get this functionality implemented in the existing artifact tool so at this point I think the Python pieces here are code-complete (we don't want to merge until we release a version of artifact tool with at least Pipeline Artifact tool support and possibly file container support).

File: azure-devops/azext_devops/dev/common/const.py [hunk 1]
```diff
@@ -24,3 +24,35 @@
 DEVOPS_TEAM_PROJECT_DEFAULT = 'project'
 PAT_ENV_VARIABLE_NAME = CLI_ENV_VARIABLE_PREFIX + 'PAT'
 AUTH_TOKEN_ENV_VARIABLE_NAME = CLI_ENV_VARIABLE_PREFIX + 'AUTH_TOKEN'
+
+# policy constants
```
atbagga: This seems like unintended change introduced with resolving merge conflict? Policy constants section should be removed from here.

azure-devops/azext_devops/dev/pipelines/_help.py
```diff
@@ -48,3 +48,15 @@ def load_pipelines_help():
...
+
+    helps['pipelines runs'] = """
+    type: group
+    short-summary: Manage pipeline runs.
+    long-summary:
+    """
+
+    helps['pipelines runs artifact'] = """
+    type: group
+    short-summary: Download, upload and list run artifacts.
```
atbagga: Short summary: "Manage pipelines artifacts" should be enough so we do not have to revisit the help when we add more commands under the artifact subgroup and consistent with other summary help texts.

azure-devops/azext_devops/dev/pipelines/commands.py
```diff
@@ -43,6 +43,14 @@
     exception_handler=azure_devops_exception_handler
 )
 
+runOps = CliCommandType(
```
atbagga: defined twice next to each other

azure-devops/azext_devops/dev/pipelines/commands.py
```diff
@@ -77,3 +85,8 @@ def load_build_commands(self, _):
         # basic release commands
         g.command('list', 'release_definition_list', table_transformer=transform_release_definitions_table_output)
         g.command('show', 'release_definition_show', table_transformer=transform_release_definition_table_output)
+
+    with self.command_group('pipelines runs artifact', command_type=runOps) as g:
```
gauravsaralMs: I am a bit confused on this..
This command can download build artifacts as well?
If yes then should there be `pipelines build runs artifact` command group as well?
c: @geverghe  for inputs

azure-devops/azext_devops/dev/pipelines/runs.py
```diff
@@ -0,0 +1,67 @@
+# --------------------------------------------------------------------------------------------
```
gauravsaralMs: This pipelines folder is for pipelines vertical.
My understanding says that pipeline vertical currently has build, release and pipelines (preview)
Runs as a concept exist across all 3 (build, release, pipelines) so we might need to names these files accordingly to avoid confusion like this should be pipeline_runs. 
* If my understanding about build, release and pipeline being different is correct

azure-devops/azext_devops/dev/pipelines/commands.py
gauravsaralMs: I am waiting mainly for this to close

azure-devops/azext_devops/dev/pipelines/runs.py
atbagga: That is correct pipelines_runs would be a better name for this file.

azure-devops/azext_devops/dev/common/const.py
mitchdenny: Will do - yeah - I rebased onto master before pushing from memory so maybe I introduced that then.

azure-devops/azext_devops/dev/pipelines/commands.py
mitchdenny: @gauravsaralMs the way I've implemented it, it doesn't matter whether it is a pipeline or a build (they are the same APIs anyway right?). Is it worth implementing ```pipelines build runs artifact``` though? Seems redundant since they would be identical.

azure-devops/azext_devops/dev/pipelines/runs.py
mitchdenny: OK - I'll rename to ```pipelines_runs.py``` in the next push. Would be good to get clarity on whether we need to have ```pipelines runs artifacts``` AND ```pipelines build runs artifacts```. To me it seems like unnecesary redundancy.

azure-devops/azext_devops/dev/pipelines/commands.py
gauravsaralMs: agree @mitchdenny . internally both can be same.
but the question is more around "is it build artifact and pipeline artifact same from a user perspective ?".
cc: @geverghe

geverghe: My apologies for the delayed response. As @gauravsaralMs pointed out, `pipelines runs artefact` and `build artifact` are two different concepts segregated from a user experience/concept. Although behind the scenes we are using the same build API, it is a temporary fix till the pipelines APIs are out. Therefore in summary: 
1. `pipelines runs artifact` and `build artifact` (no runs here since build and build-definition are separate sub groups) are different conceptually. 
2. It is worth implementing `pipelines build runs artifact` and switch it out to the pipelines API once it is done.

mitchdenny: We've had a side conversation with the Pipelines team about this. The net result is that we are going to just have the one ```runs``` command tree under ```pipelines```. The unified pipeline experience is arriving in the very short term and there is no point adding additional ```runs``` command trees when they are going to be relatively short lived - especially since this is the first CLI download experience we have for Pipeline-based artifacts.

azure-devops/azext_devops/dev/pipelines/runs.py
mitchdenny: In light of the conversations with the Pipelines team I am proposing that we keen the filename as ```runs.py``` and not rename it to ```pipeline_runs.py``` given that there will be only one ```runs``` command tree. I'm going to resolve this for now since I think that is pretty self evident but if there is another reason to rename to ```pipeline_runs.py``` reopen this comment.

mitchdenny:
Hey @atbagga, @gauravsaralMs  and @geverghe - I've responded to the PR feedback. A few renames etc didn't happen following a conversation with Pipelines on the right course of action, but a few of the other feedback items were addressed.

I still don't want to merge just yet, need to work with Artifacts Core to flight the build of ArtifactTool that can support Pipeline Artifact download (0.2.99). We'll flight an update which supports File Containers too in the near future but that can happy after this is merged.

mitchdenny:
@zarenner this is a largish change I was working on prior to you doing your distro fixes. I've merged your changes in and moved things around a bit. It all seems to test out OK but wanted to you be aware of the change in where artifact tool-related modules have been moved to (moved because we will also be using them for ```pipelines runs artifact``` command-tree.

mitchdenny:
> Hey @atbagga, @gauravsaralMs  and @geverghe - I've responded to the PR feedback. A few renames etc didn't happen following a conversation with Pipelines on the right course of action, but a few of the other feedback items were addressed.
> I still don't want to merge just yet, need to work with Artifacts Core to flight the build of ArtifactTool that can support Pipeline Artifact download (0.2.99). We'll flight an update which supports File Containers too in the near future but that can happy after this is merged.

OK - the version of ArtifactTool that supports downloading Pipeline Artifacts has been flighted. This is ready to merge when the Protocol team is ready :)

atbagga:
@geverghe  We might end up doing one more release with 4.0 client. i.e. without the new pipelines commands I am working on. If we are ok to release these commands under pipelines runs artifact list/download/upload. We should be good to merge this PR. 

If you want to take all the pipelines commands together we have to hold this PR merge. 

My vote would be that we merge and release this. :)

geverghe:
@atbagga  - I am ok with merge and release of the artefact commands. However, we need to bash prior to release. Let's schedule some time.
@snankani - FYI.

mitchdenny:
Let's merge it :) I just did a test outside of my dev environment by downloading the WHL file and its working fine but I agree a bug bash would be good.

atbagga:
@gauravsaralMs  @ishitam8  @snankani 
PR merge is blocked for second approver. Please take a look.

PR #467, opened 2019-02-28
Changes to make sure back compatibility across extension versions is not broken

gauravsaralMs:
There are 2 major changes:
1. Run tests from old release branch with new code, which will make sure that old scripts work with new extension (this will give decent coverage)
2. Using a script check following
   a. No new mandatory parameter is coming in for a command
   b. No parameter is removed from a command
Because these 2 things can break back compatibility

https://github.com/Microsoft/azure-devops-cli-extension/issues/462

### DIFF
--- a/scripts/backCompatChecker.py
+++ b/scripts/backCompatChecker.py
@@ -19,8 +19,8 @@ class Arguments(dict):
 
 def extractArgumentsFromCommand(command):
     argumentList = []
-    commandExtended = 'cmd /c az ' + command + ' -h'
-    help_text = subprocess.run(commandExtended.split(' '),stdout=subprocess.PIPE)
+    commandExtended = 'az ' + command + ' -h'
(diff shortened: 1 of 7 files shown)


scripts/backCompatChecker.py
```diff
@@ -0,0 +1,115 @@
...
+        self.isRequired = isRequired
+        dict.__init__(self,command = command, name = name, isRequired = isRequired)
+
+def extractArgumentsFromCommand(command):
+    print('running extractArgumentsFromCommand for ' + command)
+    argumentList = []
+    commandExtended = 'az ' + command + ' -h'
+    help_text = subprocess.run(commandExtended.split(' '), shell=True, stdout=subprocess.PIPE)
+    print('help text for ' + command)
+    print(help_text)
```
ishitam8: You can remove these print statements

gauravsaralMs: keeping this because it will help with debugging, specially with help text from released extension so that developer will not have to do that manually again

Issue #102, opened 2017-11-30
Manage Users


jadiaz:
CLI should have the ability to add/remove users (Basic, Stakeholders)

EmFl:
Hello, 
would it be possible to have some info on when this is going to be made available in cli? 
Thanks!

geverghe:
@EmFl we are working on a few other items and should be able to get to this by Feb, 2019 as per plan. I will provide an update in case there is a change in plan. Feel free to make a contribution and we can guide you in case you run into any issues.

geverghe:
Here is the spec against which we are actively developing the user management commands:

| Scenario | Command | Comments/Questions|
|--|--|--|
| As an admin, I want to add a new user to my organization| `az devops user add --user-id` | 1. Why is the verb add when we have had create everywhere? <br> Since this is not actually creating a user, the verb now gives a wrong connotation to intent. It affects uniform commanding but making it create threatens intent. Need to do additional validation on this.| 
| As an admin, I want to remove a user from my organization.| `az devops user remove --user-id` | |
| As an admin, I want to view all users in the organization| `az devops user list` | |
| As an admin, I want to update the access version for a user | `az devops user update --user-id --access-level` | |
| As an admin, I want to view the details pertaining to a user | `az devops user show --user-id ` | |

Would love to hear your thoughts on this

ishitam8:
This is available in version 0.4.0. Hence, closing this.

PR #484, opened 2019-03-12
Commands for managing users in an org

ishitam8:

 - [ ] : Approach is signed off on the issue. #102


azure-devops/azext_devops/dev/team/arguments.py
```diff
@@ -20,6 +20,7 @@
 _SERVICE_ENDPOINT_TYPE = [SERVICE_ENDPOINT_TYPE_GITHUB, SERVICE_ENDPOINT_TYPE_AZURE_RM]
 _SERVICE_ENDPOINT_AUTHORIZATION_SCHEME = [SERVICE_ENDPOINT_AUTHORIZATION_PERSONAL_ACCESS_TOKEN,
                                           SERVICE_ENDPOINT_AUTHORIZATION_SERVICE_PRINCIPAL]
+_ACCESS_LEVELS = ['advanced', 'earlyAdopter', 'express', 'none', 'professional', 'stakeholder']
```
gauravsaralMs: Is this a fixed list? More specifically is this expected to change in near future?
Is there any way we can query this from some rest api

azure-devops/azext_devops/dev/team/arguments.py
```diff
@@ -51,6 +52,13 @@ def load_team_arguments(self, _):
         context.argument('use_git_aliases', **enum_choice_list(_YES_NO_SWITCH_VALUES))
         context.argument('list_config', options_list=('--list', '-l'))
 
+    with self.argument_context('devops user') as context:
+        context.argument('access_level', **enum_choice_list(_ACCESS_LEVELS))
```
gauravsaralMs: let use `arg_type=get_enum_type` for consistency

azure-devops/azext_devops/dev/team/user.py
```diff
@@ -0,0 +1,108 @@
...
+                                                                                          GraphUser,
+                                                                                          JsonPatchOperation)
+from azext_devops.dev.common.services import (get_member_entitlement_management_client,
+                                              resolve_instance)
+from azext_devops.dev.common.identities import resolve_identity_as_id
+
+
+def get_user_entitlements(top=None, skip=None, organization=None, detect=None):
+    """List users for an organization.
+    :param int top: Maximum number of the users to return. Max value is 10000. Default value is 100
```
gauravsaralMs: If we put `top=100` instead of `top=None` then knack will take care of this in the standard format
maybe we can use that

azure-devops/azext_devops/dev/team/_format.py
```diff
@@ -122,6 +122,35 @@ def _transform_team_member_row(row):
...
+
+
+def _transform_user_row(row):
+    table_row = OrderedDict()
+    table_row['ID'] = row['id']
+    if 'user' in row:
+        table_row['Display Name'] = row['user']['displayName']
+        table_row['Email'] = row['user']['mailAddress']
+    else:
+        table_row['Display Name'] = ' '
```
gauravsaralMs: why are we not setting `table_row['Email']` = '' in this case?

azure-devops/azext_devops/dev/team/user.py
```diff
@@ -0,0 +1,108 @@
...
+    value = {}
+    value['accessLevel'] = user_access_level
+    value['extensions'] = []
+    value['projectEntitlements'] = []
+    value['user'] = graph_user
+    patch_document = []
+    patch_document.append(_create_patch_operation('add', '', value))
+    do_not_send_invite = False
+    if send_email_invite != 'true':
+        do_not_send_invite = True
```
gauravsaralMs: can we use `resolve_true_false` from common.arguments?

azure-devops/azext_devops/dev/team/_format.py
atbagga: Also just for my knowledge when would it happen that the email, access level, and status information will be missing?

azure-devops/azext_devops/dev/team/arguments.py
atbagga: Should we keep the input options like.. early-adopter instead of earlyadopter? 
@geverghe  PM choice :)

azure-devops/azext_devops/test/team/test_user.py
```diff
@@ -0,0 +1,103 @@
...
+from azext_devops.dev.team.user import (get_user_entitlements,
+                                        get_user_entitlement,
+                                        add_user_entitlement,
+                                        delete_user_entitlement,
+                                        update_user_entitlement)
+
+from azext_devops.dev.common.services import clear_connection_cache
+
+    
+class TestUserMethods(unittest.TestCase):
```
atbagga: Check the change in this PR to refactor authentication mocks into a base class. https://github.com/Microsoft/azure-devops-cli-extension/pull/469

Can you move this to common place and consume in your tests.

azure-devops/azext_devops/dev/team/arguments.py
geverghe: For options, we have stuck to non hyphenated values - not sure if this was intentional but it feels easier to copy paste. For parameters we have followed hyphenated word separation. I would stick to non-hyphenated version for consistency. Are there any other challenges you foresee?

atbagga: No. That sounds good.

ishitam8: Yes, we can query this from [REST API](https://docs.microsoft.com/en-us/rest/api/azure/devops/memberentitlementmanagement/user%20entitlement%20summary/get?view=azure-devops-rest-5.0) (which returns list of valid licenses/access-level for that account)  . The list seems to be fixed though, that's why adding new command was avoided.
@geverghe

gauravsaralMs: If we know this is not going to change in near future than hardcoding is fine, else we should get this from rest api.

azure-devops/azext_devops/dev/team/_format.py
ishitam8: That check was for principalName. Switched to mail address, that's why not required now. Thanks

azure-devops/azext_devops/test/team/test_user.py
ishitam8: Sure, will refactor this along with Group management commands /tests.

PR #489, opened 2019-03-19
Support debugging from VSCode

gauravsaralMs:

### DIFF
--- a/scripts/run_az.py
+++ b/scripts/run_az.py
@@ -1,0 +1,10 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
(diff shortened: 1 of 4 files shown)


scripts/run_az.py
```diff
@@ -0,0 +1,13 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import sys
+
+from azure.cli import __main__ as cli_main
+
+# If import here is failing then it means you need to install azure cli core in machine
```
atbagga: Handle import failure and throw this is exception message?

gauravsaralMs: No... I want use to see the exact line of failure and this is supposed to run in debug mode so try..handle... is actually not going to add any value

Issue #471, opened 2019-03-04
Manage security groups


geverghe:

geverghe:
Spec against which we are actively developing the security group management commands:

## Security Group Management
| Scenario | Command | Comments |
|--|--|--|
| As an admin, I want to create a security group| `az devops security group create --name --description --project ` | |
| As an admin, I want to update a security group details| `az devops security group create --group-id --name --description --project ` | |
| As an admin, I want to delete a security group| `az devops security group delete --group-id`| |
| As an admin, I want to view all security groups in project | `az devops security group list --project`| |

## Security Group Membership Management
| Scenario | Command | Comments |
|--|--|--|
| I want to add a member to the security group | `az devops security group membership add --member-id --group-id`| |
| I want to add the security group as a member of another group| `az devops security group membership add --member-of-id --group-id`| |
| I want to remove a member from the security group | `az devops security group membership remove --member-id --group-id`||
| I want to remove a the security group as a member of another security group | `az devops security group membership remove --member-of-id --group-id`||
| I want to view all members of the security group | `az devops security group membership list` or `az devops security group membership list --relationship member`|
| I want to view the security groups to which the group belongs to | `az devops security group membership list --relationship memberof` | |

Ashu6753:
Hi geverghe,

By when should we expect this feature to be added to the azure devOps cli? Is there any timeline for this?
Also is there any way we can assign extension to a user account through azure-devops-cli.
Looking forward for your reply. Thanks in advance

geverghe:
Hey @Ashu6753 , thank you for reaching out. We are looking at this coming out around mid of April. I will keep you posted in case there are any change of plans. We haven't factored the extension assignment in user management today. Could you raise a feature request for this? This will help us hear from others with similar needs and help us prioritize.

Ashu6753:
> Hey @Ashu6753 , thank you for reaching out. We are looking at this coming out around mid of April. I will keep you posted in case there are any change of plans. We haven't factored the extension assignment in user management today. Could you raise a feature request for this? This will help us hear from others with similar needs and help us prioritize.

Hi geverghe,

Thanks for the response. I have created feature request #485 for the extension assignment in user management.

Extension Assignment in user management[Feature Request] #485

ishitam8:
Updating all the command structures to use descriptors instead of UID.
[...]

PR #491, opened 2019-03-20
Group and membership management commands

ishitam8:

 - [ ] : This PR has a corresponding issue open in the Repository. #471 


azure-devops/azext_devops/dev/team/security_group.py
```diff
@@ -0,0 +1,202 @@
...
+    scope_descriptor = None
+    if project is not None:
+        project = get_project_id_from_name(organization, project)
+        scope_descriptor = get_descriptor_from_storage_key(project, client)
+    if subject_types is not None:
+        subject_types = subject_types.split(',')
+    group_list_response = client.list_groups(scope_descriptor=scope_descriptor,
+                                             continuation_token=continuation_token, subject_types=subject_types)
+    if group_list_response.continuation_token is not None:
+        print('Showing only 500 groups. ' +
```
atbagga: This is not the best experience for automation. You removed the continuation token from the json response. 
It is ok to print it for table output but it should not be removed from json.

azure-devops/azext_devops/dev/team/security_group.py
```diff
@@ -0,0 +1,202 @@
...
+              'To list next set of groups use this token as --continuation-token argument and run the command again.' +
+              ' TOKEN:', group_list_response.continuation_token)
+    return group_list_response.graph_groups
+
+
+def create_group(name=None, description=None, origin_id=None, email_id=None,
+                 groups=None, project=None, organization=None, detect=None):
+    """Create a new Azure DevOps group or materialize an existing AAD group.
```
gauravsaralMs: this is debatable:
create vs materialize look like 2 different operations and maybe should have 2 different commands..

gauravsaralMs: like, can I use `groups` when materializing AAD group?

azure-devops/azext_devops/dev/team/security_group.py
ishitam8: Moved this to table formatting and retained the json response

azure-devops/azext_devops/dev/team/security_group.py
```diff
@@ -0,0 +1,202 @@
...
+    :param str project: List groups for a particular project.
+    :param str continuation_token : If there are more results than can't be returned in a single page, the result set
+                                    will contain a continuation token for retrieval of the next set of results.
+    :param [str] subject_types: A comma separated list of user subject subtypes to reduce the retrieved results.
+    """
+    organization = resolve_instance(detect=detect, organization=organization)
+    client = get_graph_client(organization)
+    scope_descriptor = None
+    if project is not None:
+        project = get_project_id_from_name(organization, project)
```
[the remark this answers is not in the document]
ishitam8: We will support both name and project_id

azure-devops/azext_devops/dev/team/security_group.py
```diff
@@ -0,0 +1,202 @@
...
+    group_list_response = client.list_groups(scope_descriptor=scope_descriptor,
+                                             continuation_token=continuation_token, subject_types=subject_types)
+    if group_list_response.continuation_token is not None:
+        print('Showing only 500 groups. ' +
+              'To list next set of groups use this token as --continuation-token argument and run the command again.' +
+              ' TOKEN:', group_list_response.continuation_token)
+    return group_list_response.graph_groups
+
+
+def create_group(name=None, description=None, origin_id=None, email_id=None,
```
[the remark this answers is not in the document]
ishitam8: Yep, examples would be helpful here. I'll add some, specially for AAD groups

azure-devops/azext_devops/dev/team/security_group.py
ishitam8: Yes, i can use groups when materializing AAD group. I don't think adding a separate command would add much value, since functionality would be almost same as this one. 
Don't you think adding examples would be enough instead of a separate command ?
@geverghe  Thoughts?

geverghe: @gauravsaralMs - I think it's fair - it's fundamentally creating a group in Azure DevOps. IMO, the rest are implementation nitty gritty. The reasoning is as follows:
Can I create a group in Azure DevOps using cli? Yes, use group command.
Can I create a group for an AAD group in Azure DevOps? Yes, use group command.
Do you see a disconnect in the message?

ishitam8: I have added few examples for create command. @gauravsaralMs  and @geverghe  Let me know if this would be useful enough to consume this command.

azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -38,6 +38,35 @@ def load_team_help():
...
+    type: command
+    short-summary: Create a new Azure DevOps group or materialize an existing AAD group.
+    long-summary:
+    examples:
+          - name: Create an Azure DevOps Group with name and description
+            text: |
+              az devops security group create --name 'Some group name' --description
+              'Something to describe this group' 
+              
+          - name: Materialize an existing AAD Group with its origin ID/AAD object ID
```
geverghe: IMO, materialize seems to be a confusing terminology or is it an understood terminology? Is there any documentation around the same that uses "materialize" as the terminology?

azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -38,6 +38,35 @@ def load_team_help():
...
+            text: |
+              Get object ID of the existing AAD group 
+              az ad group show -g {Group Name}
```
geverghe: Is it the group name or the descriptor?

azure-devops/azext_devops/dev/team/security_group.py
```diff
@@ -0,0 +1,222 @@
...
+                                    GraphGroupOriginIdCreationContext)
+
+logger = get_logger(__name__)
+
+
+def list_groups(project=None, continuation_token=None, subject_types=None, organization=None, detect=None):
+    """ List all groups.
+    :param project: List groups for a particular project.
+    :type project: str
+    :param continuation_token : If there are more results than can't be returned in a single page, the result set
```
gauravsaralMs: tha**t** can't be

azure-devops/azext_devops/dev/team/security_group.py
gauravsaralMs: not sure about users, but looked a bit confusing to me. 
I am fine with any approach

azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -38,6 +38,35 @@ def load_team_help():
...
+              az devops security group create --origin-id {Object ID}
+              Use this descriptor to add this group to other Azure DevOps Groups.
+
+          - name: Materialize an existing AAD Group with its origin ID/AAD object ID and also add it to an AzureDevOps group.
```
geverghe: This should be limited to "Add an existing AAD group to an Azure DevOps group" - remaining including saying it is done with origin ID etc are all implementation details. Focus should be on the scenario.

azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -38,6 +38,35 @@ def load_team_help():
...
+            text: |
+              Get object ID of the existing AAD group               
+              az ad group show -g {Group Name}
+              az devops security group create --origin-id {Object ID} --groups 'vssgp.someDescriptorForGroup'
```
geverghe: Have an example on bulk add as well - adding it to groups at the same time - what is the accepted method.

azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -38,6 +38,35 @@ def load_team_help():
...
+
+          - name: Materialize an existing AAD Group with its Email ID
```
geverghe: Knock off this scenario as this is not something we are focusing on. We are focusing on adding AAD group to existing security groups. We need not bother about materializing AAD groups in Azure DevOps just yet as a scenario.

azure-devops/azext_devops/dev/team/_help.py
ishitam8: "Add an existing AAD group to an Azure DevOps group"  This looks confusing when you compare it to 'membership add command' where we are supporting scenario of addition of group to other group.

azure-devops/azext_devops/dev/team/security_group.py
ishitam8: Removing any context related to 'materializing a group' , since its internal to implementation details

azure-devops/azext_devops/dev/team/_help.py
ishitam8: No, there is no documentation on the terminology

azure-devops/azext_devops/dev/team/_help.py
ishitam8: Removing any context related to 'materializing a group' , since its internal to implementation details

ishitam8:
> we should also plan, how can we run recording tests on this because this is quite complicated to test regularly

Yep, i think we would be able to test most of the scenarios with recording tests except for AAD group addition/removal

Issue #436, opened 2019-02-13
package name "vstsCompressed" does not follow the python convention of naming


atbagga:
It should be all lowercase name. Either **vsts** or **vsts_compressed**.

Issue #416, opened 2019-02-05
Remove dependency from 4.0 azure devops client


gauravsaralMs:
https://github.com/Microsoft/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/dev/common/services.py 

All v4_0 should be moved to v4.1

snankani:
I feel we should move to latest version once we get the new sdk for pipelines commands.

gauravsaralMs:
> I feel we should move to latest version once we get the new sdk for pipelines commands.

makes sense
fyi: @atbagga as we will need to align this with pipeline work

atbagga:
#492

gauravsaralMs:
@atbagga : Should we close this now?

PR #492, opened 2019-03-20
[Major Change] Moving API version to 5.x

atbagga:
Tracking feature: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1481658

Fixes: #436, #416 

Adding a task to re-enabled the pipeline job to run tests against older release. #510


azure-devops/azext_devops/dev/common/services.py
```diff
@@ -136,93 +136,93 @@ def get_token_from_az_login(profile, user, tenant):
...
-    for key in _vss_connection:
+    for key in _connection:
         return key
 
 
 def get_release_client(team_instance=None):
-    connection = get_vss_connection(team_instance)
-    return connection.get_client(VSTS_MODULE + 'release.v4_0.release_client.ReleaseClient')
+    connection = get_connection(team_instance)
+    return connection.get_client(VSTS_MODULE + 'v5_0.release.release_client.ReleaseClient')
```
gauravsaralMs: In latest python client there is a concept of [released client factory](https://github.com/Microsoft/azure-devops-python-api/blob/dev/azure-devops/azure/devops/released/client_factory.py), can we use that instead of hard coding it here now?

atbagga: Released folder does not have all the clients. 

On using ClientFactory_v50 the issue is it tries to get the clients via 'azure.devops.v5_0' but since we are not using the pip package this again falls into the namespace clash where we will not get devops in azure.

gauravsaralMs: we should fix the function in factory in that case.
If we can't take it right now then also please open a task because sooner or later we should move to use this

azure-devops/azext_devops/dev/common/exception_handler.py
```diff
@@ -12,7 +12,7 @@
 def azure_devops_exception_handler(ex):
     # we are taking dependency on string here because taking dependency on
     # Azure DevOps Client will increase load time for every command
-    if type(ex).__name__ == 'VstsServiceError':
+    if type(ex).__name__ == 'AzureDevOpsServiceError':
         logger.debug('handling vsts service error')
```
gauravsaralMs: are we fixing logging as well in this PR?  'handling vsts service error' => 'handling azure devops service error'

azure-devops/azext_devops/dev/boards/work_item.py
```diff
@@ -160,21 +159,22 @@ def update_work_item(id, title=None, description=None, assigned_to=None, state=N
...
     try:
-        organization = resolve_instance(detect=detect, organization=organization)
+        organization, project = resolve_instance_and_project(detect=detect, organization=organization, project=project)
         client = get_work_item_tracking_client(organization)
-        delete_response = client.delete_work_item(id, destroy)
+        delete_response = client.delete_work_item(id=id, project=project, destroy=destroy)
         print('Deleted work item {}'.format(id))
         return delete_response
-    except VstsServiceError as ex:
+    except AzureDevOpsServiceError as ex:
```
gauravsaralMs: note to self: fix these as these are not needed anymore 

cc: @gauravsaralMs

azure-devops/azext_devops/dev/common/exception_handler.py
atbagga: No, this one was just functional changes anything that was caught in the tests.

But I will take any non functional changes like this in a follow up Pr.

azure-devops/azext_devops/dev/common/services.py
atbagga: We can but more than the upside of removing hardcoding from our code, It has a downside of not having the freedom to replace the package content with the latest. I will open a task anyways for using the client factory.

atbagga: #509 Filed a task marking this as resolved.

PR #496, opened 2019-03-25
Recorded tests for User commands

ishitam8:

 - [ ] : This PR has a corresponding issue open in the Repository.  : Few bug bash fixes
 - [ ] : Approach is signed off on the issue. NA

Adding recorded tests for User related commands and below fixes.
1. Updated help text for user list command
2. Updated exception in case of invalid license type.


azure-devops/azext_devops/dev/team/user.py
```diff
@@ -67,8 +68,11 @@ def update_user_entitlement(user, license_type, organization=None, detect=None):
     if '@' in user:
         user = resolve_identity_as_id(user, organization)
     client = get_member_entitlement_management_client(organization)
-    user_entitlement_update = client.update_user_entitlement(document=patch_document, user_id=user)
-    return user_entitlement_update.user_entitlement
+    try:
+        user_entitlement_update = client.update_user_entitlement(document=patch_document, user_id=user)
+        return user_entitlement_update.user_entitlement
+    except Exception:
```
gauravsaralMs: not very sure if we should do a blanket catch and convert it to 'Invalid license type.'.
I mean that the exception here can be anything (maybe a bug on service side)

azure-devops/azext_devops/dev/team/user.py
```diff
@@ -13,7 +14,7 @@
 
 
 def get_user_entitlements(top=100, skip=None, organization=None, detect=None):
-    """List users for an organization.
+    """List users for an organization [except for AAD users].
```
gauravsaralMs: I think this will show AAD users also if user is directly added in Users tab, but this does not expand AAD group.
So we should tell that, the current text looks incorrect

tests/test_devopsUserTest.py
```diff
@@ -0,0 +1,71 @@
...
+    @disable_telemetry
+    @set_authentication
+    def test_devops_user_command_addUpdateListShowRemove(self):
+        self.cmd('az devops configure --defaults organization=' +  DEVOPS_CLI_TEST_ORGANIZATION)
+        try:
+            # check user list before adding the user
+            user_list_response = self.cmd('az devops user list -o json --detect off').get_output_in_json()
+            assert len(user_list_response) > 0
+            user_id_found = False
+            for item in user_list_response :
```
gauravsaralMs: this looks a bit incorrect, what if my test failed once at some point after adding user (then it will fail again)
I think first we should detect and take the org to an expected starting point (where this user is not added)

tests/test_devopsUserTest.py
```diff
@@ -0,0 +1,71 @@
...
+                if item['user']['mailAddress'] == _TEST_EMAIL_ID:
+                    user_id_found = True
+            assert user_id_found == False
+
+            #add user
+            user_add_response = self.cmd('az devops user add -o json --detect off --user ' + _TEST_EMAIL_ID + ' --license-type stakeholder').get_output_in_json()
```
gauravsaralMs: can we add early adopter here also?
That way this test will be able to run on non-microsoft orgs as well?

azure-devops/azext_devops/dev/team/user.py
ishitam8: Currently, there is an issue in deserialization of response in the client. I''ll raise a bug, and once that fix goes in the client, i shall update the exception logic in our code.

azure-devops/azext_devops/dev/team/user.py
ishitam8: "List users for an organization **[except for users which are added via AAD groups]** "
Does this makes sense @gauravsaralMs  , @geverghe ?

tests/test_devopsUserTest.py
ishitam8: Makes sense, i'll update this.

tests/test_devopsUserTest.py
ishitam8: EarlyAdopter can't be assigned manually for microsoft account. And other license types won't be updated for microsoft account. Hence, removing the license types related checks so that these tests run successfully for all accounts/users.

azure-devops/azext_devops/dev/team/user.py
atbagga: In that case should we throw some generic exception rather than a specific exception which could be misleading?

ishitam8: Yeah, but then user won't know what the issue is. I think this is fine as its a temporary and once the actual issue is fixed in python client we can throw a specific error in case of invalid access level and a generic one otherwise.

PR #499, opened 2019-03-25
Refactoring authentication code for UTs

ishitam8:

### DIFF
--- a/azure-devops/azext_devops/test/utils/authentication.py
+++ b/azure-devops/azext_devops/test/utils/authentication.py
@@ -15,19 +15,19 @@ from .helper import UNIT_TEST_PAT_TOKEN
 class AuthenticatedTests(unittest.TestCase):
 
     def authentication_setUp(self):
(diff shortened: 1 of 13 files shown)


azure-devops/azext_devops/test/utils/authentication.py
```diff
@@ -0,0 +1,30 @@
...
+    # Attempt to load mock (works on Python 3.3 and above)
+    from unittest.mock import patch
+except ImportError:
+    # Attempt to load mock (works on Python version below 3.3)
+    from mock import patch
+from .helper import UNIT_TEST_PAT_TOKEN
+
+class AuthenticatedTests(unittest.TestCase):
+
+    def authentication_setUp(self):
```
atbagga: nit: authentication_set**u**p

ishitam8:
> Why do we need to authenticate in each test instead of doing it in the setUp function itself?

Because it's not required in for every test case in every test. But I'll move it to setUp wherever it is used in all the tests.

PR #501, opened 2019-03-26
Fix live test because service changed response which it was sending, …

gauravsaralMs:
Fix live test because service changed response which it was sending null, instead of sending false for a value it is now sending empty


gauravsaralMs:
> Is this not considered as breaking change in API?

it's documented as object, it remains an object so not breaking change

PR #507, opened 2019-03-27
Allow support to break back compat when it is required

gauravsaralMs:

This is needed to allow 
https://github.com/Microsoft/azure-devops-cli-extension/pull/506

### DIFF
--- a/scripts/backCompatChecker.py
+++ b/scripts/backCompatChecker.py
@@ -9,7 +9,8 @@ import subprocess
 
 oldArguments = []
 newArguments = []
-allowedMissingArguments = ['--azure-rm-service-prinicipal-key']
+allowedMissingArguments = {}
+allowedMissingArguments['devops service-endpoint'] = ['--azure-rm-service-prinicipal-key']
 
 class Arguments(dict):
     def __init__(self, command, name, isRequired):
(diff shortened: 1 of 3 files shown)


scripts/backCompatChecker.py
```diff
@@ -9,6 +9,7 @@
 
 oldArguments = []
 newArguments = []
+allowedMissingArguments = ['--azure-rm-service-prinicipal-key']
```
atbagga: This is prone to letting issues creep in. 
How would we solve this is the parameter is generic like "project" name? If I ignore that it will be ignored for all commands.

Regarding the PR for 5.0 upgrade. I would rather not want to add exceptions here for long term. We know whatever is breaking in this major upgrade. If I add checks for those here, after one release those checks should not be applied.

atbagga: We can use a dictionary to specifically allow exceptions for params corresponding to each command. e.g. { command1: [arg1, arg2], command2: [arg1, arg3, arg4] }

I can use the same for 5.0 unless you have a better solution in mind. I was thinking of disabling this for one release since we are introducing breaking changes but then this looks like a good way to catch all the breaking changes, document them in release notes and then get rid of the checks next time.

scripts/backCompatChecker.py
```diff
@@ -108,7 +110,13 @@ def findExtension():
             break
 
     if isArgumentMissing is True:
-        errorList.append('Argument missing for command ' + oldArgument.command + ' argument ' +  oldArgument.name)
+        allowedEntryFound = False
```
atbagga: nit: If block can be simplified to (do not need to reiterate the items since it is a dictionary) -

if not (allowedMissingArguments.get(oldArgument.command) and oldArgument.name in allowedMissingArguments.get(oldArgument.command)):
errorList.append('Argument missing for command ' + oldArgument.command + ' argument ' +  oldArgument.name)

PR #512, opened 2019-03-28
remove ps1 scripts

gauravsaralMs:


atbagga:
I use runTests.ps1 to run tests locally since it takes care of deploy and run the tests. 
Now I will have to deploy latest code myself and then run tests using "pytest". 

This is sub-optimal but fine. 

**If we want to take the change as is -** this will need to go into the dev_setup documentation since it mentions runTests.ps1. 

OR 

We can take the pipeline changes but leave the script as a utility. (If this is to get rid of powershell scripts from dev setup we can convert the same to python for which there is a partial PR in progress. I stopped working since lot of pipeline changes went in parallel to that making the changes outdated.) 

Thoughts?

gauravsaralMs:
> I use runTests.ps1 to run tests locally since it takes care of deploy and run the tests.
> Now I will have to deploy latest code myself and then run tests using "pytest".
> 
> This is sub-optimal but fine.
> 
> **If we want to take the change as is -** this will need to go into the dev_setup documentation since it mentions runTests.ps1.
> 
> OR
> 
> We can take the pipeline changes but leave the script as a utility. (If this is to get rid of powershell scripts from dev setup we can convert the same to python for which there is a partial PR in progress. I stopped working since lot of pipeline changes went in parallel to that making the changes outdated.)
> 
> Thoughts?

so now UTs can run against latest code from VSCode itself.
For live test installing extension step is involved but that is also Ctrl+Shift+B ==> select option 
and that is how most of the tests are which are not UTs exactly.

I am inclined towards a future where most (if not all) the dev work can be done from IDE instead of depending on scripts

atbagga:
> > I use runTests.ps1 to run tests locally since it takes care of deploy and run the tests.
> > Now I will have to deploy latest code myself and then run tests using "pytest".
> > This is sub-optimal but fine.
> > **If we want to take the change as is -** this will need to go into the dev_setup documentation since it mentions runTests.ps1.
> > OR
> > We can take the pipeline changes but leave the script as a utility. (If this is to get rid of powershell scripts from dev setup we can convert the same to python for which there is a partial PR in progress. I stopped working since lot of pipeline changes went in parallel to that making the changes outdated.)
> > Thoughts?
> 
> so now UTs can run against latest code from VSCode itself.
> For live test installing extension step is involved but that is also Ctrl+Shift+B ==> select option
> and that is how most of the tests are which are not UTs exactly.
> 
> I am inclined towards a future where most (if not all) the dev work can be done from IDE instead of depending on scripts

Inclined towards UI for CLI development? :)  
For me having the functionality in vscode is good to have but having scripts for anything that runs in the CI pipelines is a must have (I hate that style check is not straight forward due to python version dependency.) 

I am fine with manual install and pytest for running tests. But I will contribute a python script which can do everything for me from command line itself.
You will still need to update the documentation to reflect the deletion of the script.

PR #525, opened 2019-04-08
Removing task definition list and show commands 

atbagga:
Removing task definition list and show commands since the APIs are no longer publicly supported.


gauravsaralMs:
[thinking out loud]:
can we support this kind of things (when back compat is breaking and API is not very complex) using command structure similar to invoke?
Invoke will bypass the client requirement so we can still make it work 
cc: @atbagga ?

atbagga:
> [thinking out loud]:
> can we support this kind of things (when back compat is breaking and API is not very complex) using command structure similar to invoke?
> Invoke will bypass the client requirement so we can still make it work
> cc: @atbagga ?

 The decision to remove this is taken due to chances of back compat break in the API which is undocumented. 
Will forward the email thread to the team.

scripts/backCompatChecker.py
```diff
@@ -81,8 +85,9 @@ def findExtension():
 # get a set of old commands, we are not reusing the set from ext because we want to keep this clean
 oldCommands = []
 for oldArgument in oldArguments:
-    if not any(oldArgument.command in s for s in oldCommands):
-        oldCommands.append(oldArgument.command)
+    if oldArgument.command not in ignoreCommands:
+        if not any(oldArgument.command in s for s in oldCommands):
+            oldCommands.append(oldArgument.command)
```
gauravsaralMs: Exclude this from 
https://github.com/Microsoft/azure-devops-cli-extension/blob/93cae2800ffeb74e55fafbf74777bc3ed8a6913a/scripts/backCompatChecker.py#L87-L90
as well

atbagga: This was not required. 
There were other issues which I fixed.

PR #526, opened 2019-04-08
Remove repo auto detect to avoid passing repository in projectwiki create

atbagga:
If we auto detect repository we will need to pass --repository ' ' while creating project wiki else it fails. 
With this change --repository will only need to be passed with code wiki and not project wiki.

### DIFF
--- a/azure-devops/azext_devops/dev/team/wiki.py
+++ b/azure-devops/azext_devops/dev/team/wiki.py
@@ -27,7 +27,8 @@ def create_wiki(name, wiki_type='projectwiki', mapped_path=None, version=None,
     :type wiki_type: str
     :param version: [Required for codewiki type] Repository branch name to publish the code wiki from.
     :type version: str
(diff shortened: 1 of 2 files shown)


azure-devops/azext_devops/dev/team/wiki.py
```diff
@@ -33,10 +32,10 @@ def create_wiki(name, wiki_type='projectwiki', mapped_path=None, version=None,
...
     """
-    organization, project, repository = resolve_instance_project_and_repo(detect=detect,
-                                                                          organization=organization,
-                                                                          project=project,
-                                                                          repo=repository)
+    organization, project = resolve_instance_and_project(detect=detect,
+                                                         organization=organization,
+                                                         project=project,
+                                                         project_required=True)
     wiki_client = get_wiki_client(organization)
```
gauravsaralMs: can we instead make to choice to detect repo or not detect repo based on WikiType?
from a user perspective using --detect I think the expectation should remain same across commands...

PR #529, opened 2019-04-10
Fix agent version in release pipeline yml

atbagga:
Failing pipeline url: https://dev.azure.com/ms/azure-devops-cli-extension/_build?definitionId=36&_a=summary

### DIFF
--- a/.azure-pipelines/azure-pipelines-released-version.yml
+++ b/.azure-pipelines/azure-pipelines-released-version.yml
@@ -13,7 +13,7 @@ jobs:
   steps:
   - task: UsePythonVersion@0
     inputs:
-      versionSpec: '3.7.0'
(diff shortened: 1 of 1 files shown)


atbagga:
Just for information though.. Why do we have the tests running on release branch? They are not live tests so when would they fail?

gauravsaralMs:
> Just for information though.. Why do we have the tests running on release branch? They are not live tests so when would they fail?

When there is some issue with env:
example: latest azure cli is not working with our released whl
or anything like that where existing whl breaks due to some change in some other thing which we depend on

other example is 
we have dependency on package "X" (with latest) then we did a release 
then we remove dependency on package "X"
now if somehow "X" screws up (like msrest did earlier) then we will not now that released wheel is broken

PR #533, opened 2019-04-14
Update wiki page requires if match header in the request

atbagga:
Updating a wiki page requires the version to be given as if-match header. 
The client does not support it so adding it in the client for now.

This will need to be fixed in the Azure-Devops-python client. I will separately follow up on that.


atbagga:
> Just curious, how was it working earlier?
> are we adding any recording test for this?


Wasn't working. I missed testing the command since I kept it in PR for long time waiting for API version upgrade. 
I will add the tests after a bug bash for these.

Issue #93, opened 2017-11-22
vsts work item create: ability to link related work items


jdanyow:
Something like this would be super useful for our team:

```shell
vsts work item create ...required args... --parent xxxxxxxxxx
vsts work item create ...required args... --related xxxxxxxxxx
```

cc @adkinn

leeberg:
This is my #1 top request - I am using VSTS CLI to clone projects from a master template project... but without the ability to set parent / related I am left with tons of "un-parented"  tasks and PBIs with no child tasks

peteg18:
Very important for us as well!

vbossica:
@jdanyow @leeberg @peteg18 Please have a look at #125 to see if it would fit your needs.

jdanyow:
@vbossica looks good, much appreciated. Any ETA on a release?

vbossica:
@jdanyow @tedchamb knows all the details about release ETA (I'm not part of the VSTS team and don't know their release cycle)

VelinGeorgiev:
Its been 8 months since the pull request is being submitted. @tedchamb, any news on that?

tedchamb:
@VelinGeorgiev, all I can say is that there is going to be a large round of investment in the cli coming soon.  These PRs are on the list.

VelinGeorgiev:
Thanks for the update @tedchamb! Looking forward for that new update.

gauravsaralMs:
I am proposing we start with 3 commands
`az boards work-item relation add`
`az boards work-item relation remove`
`az boards work-item relation-type list`

this should solve most of the customer scenarions
also in 
`az boards work-item show` we should be getting all the fields (including relations)

Using `az boards work-item relation-type list` users can list the types of relations supported in the Org
like https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/work%20item%20relation%20types/list?view=azure-devops-rest-5.0

then they can use oppositeEndReferenceName from this command to 
 `az boards work-item relation add`
`az boards work-item relation remove`

which will look something like this 
`az boards work-item relation add --id 4 --relation System.LinkTypes.Hierarchy-Forward=1,2,3`

cc: @geverghe  @snankani  @ishitam8

geverghe:
FYI - @jdanyow , @leeberg , @peteg18 , @vbossica , @VelinGeorgiev

jboulter11:
[...]

PR #541, opened 2019-04-17
Work Item Relation Commands

gauravsaralMs:

https://github.com/Azure/azure-devops-cli-extension/issues/93

Will take up UTs in a different PR, till then recorded test will make sure nothing breaks

@geverghe : Please take a look at the help texts


azure-devops/azext_devops/dev/boards/_help.py
```diff
@@ -24,3 +24,15 @@ def load_boards_help():
     short-summary: Manage queries.
     long-summary:
     """
+
+    helps['boards work-item relation'] = """
+    type: command
```
ishitam8: **group**
For both relation and relation-type

azure-devops/azext_devops/dev/boards/arguments.py
```diff
@@ -20,3 +20,9 @@ def load_work_arguments(self, _):
     with self.argument_context('boards work-item delete') as context:
         context.argument('yes', options_list=['--yes', '-y'], action='store_true',
                          help='Do not prompt for confirmation.')
+
+    with self.argument_context('boards work-item relation') as context:
+        context.argument('id', help='The ID of the work item')
+        context.argument('relation_type', help='Relation type to create. example: parent, child ')
+        context.argument('target_ids', help='ID(s) of work-items to create relation with. \
+                         Multiple values can be passed , seperated. Example: 1,2 ')
```
ishitam8: Nit: **comma** separated

azure-devops/azext_devops/dev/boards/commands.py
```diff
@@ -27,3 +33,11 @@ def load_work_commands(self, _):
 
         # query commands
         g.command('query', 'query_work_items', table_transformer=transform_work_item_query_result_table_output)
+
+    with self.command_group('boards work-item', command_type=relationsOps) as g:
+        # relation commands
+        g.command('relation-type list', 'get_relation_types_show',
```
ishitam8: What other commands are we expecting for "relation-type' group?

gauravsaralMs: nothing as of now,, may be if we introduce relation with artifacts then we will have to do something here (not sure at this point of time)

atbagga: Then should we put it inside relation group itself?
az boards work-item relation list-types or something

gauravsaralMs: `az boards work-item relation list-types` <== I like this...
@geverghe : Thoughts?

azure-devops/azext_devops/dev/boards/relations.py
```diff
@@ -0,0 +1,107 @@
...
+
+def get_relation_types_show(organization=None, detect=None):
+    """ List work item relations supported in the organization.
+    """
+    organization = resolve_instance(detect=detect, organization=organization)
+    client = get_work_item_tracking_client(organization)
+    return client.get_relation_types()
+
+
+def add_relation(id, relation_type, target_ids, organization=None, detect=None):  # pylint: disable=redefined-builtin
```
atbagga: Table output is not very helpful. We should list all related items with relation type.
azure-devops> az boards work-item relation add --id 103 --relation-type 'Duplicate of' --target-ids 104
image

azure-devops/azext_devops/dev/boards/relations.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def get_system_relation_name(client, relation_type):
+    relation_types_from_service = client.get_relation_types()
+
+    for relation_type_from_service in relation_types_from_service:
+        if relation_type_from_service.name.lower() == relation_type.lower():
+            return relation_type_from_service.reference_name
+
+    raise CLIError("--relation-type is not valid")
```
atbagga: Might be helpful to expand this to .. 
Use az boards work-item relation-type list  command to list possible relation types in your project.

azure-devops/azext_devops/dev/boards/relations.py
```diff
@@ -0,0 +1,107 @@
...
+    work_item_query_clause = []
+    for target_work_item_id in target_work_item_ids:
+        work_item_query_clause.append('[System.Id] = {}'.format(target_work_item_id))
+
+    wiql_query_format = 'SELECT [System.Id] FROM WorkItems WHERE ({})'
+    wiql_query_to_get_target_work_items = wiql_query_format.format(' OR '.join(work_item_query_clause))
+
+    wiql_object = Wiql()
+    wiql_object.query = wiql_query_to_get_target_work_items
+    target_work_items = client.query_by_wiql(wiql=wiql_object).work_items
```
atbagga: For an id which does not exist this might become empty list. 
Here is the error I am getting which is not self explanatory. 

az boards work-item relation add --id 103 --relation-type 'Parent' --target-ids 106 -o json
**At least one operation is required for Apply.**

azure-devops/azext_devops/dev/boards/commands.py
ishitam8: Yeah, list-types sounds better in that case.

azure-devops/azext_devops/dev/boards/arguments.py
```diff
@@ -20,3 +20,9 @@ def load_work_arguments(self, _):
     with self.argument_context('boards work-item delete') as context:
         context.argument('yes', options_list=['--yes', '-y'], action='store_true',
                          help='Do not prompt for confirmation.')
+
+    with self.argument_context('boards work-item relation') as context:
+        context.argument('id', help='The ID of the work item')
+        context.argument('relation_type', help='Relation type to create. example: parent, child ')
```
atbagga: You will have to split the help text this shows up as create for both add and remove.

azure-devops/azext_devops/dev/boards/relations.py
gauravsaralMs: added a simple check, not adding a very clear one because it is an error case and not the primary use case

azure-devops/azext_devops/dev/boards/relations.py
gauravsaralMs: done

azure-devops/azext_devops/dev/boards/arguments.py
gauravsaralMs: nice catch

azure-devops/azext_devops/dev/boards/relations.py
gauravsaralMs: Relations are array so that can grow, I really don't think anyone should be trying to consume the output as table if thy want to get relations

geverghe: Can't we have the same terminology -? ID, Type Title Target ID Relationship Type? What is stopping us from providing a useful table output? Is it only the assumption that no one would use table?

geverghe: Plus - where are you getting this wheel from? 🤔

azure-devops/azext_devops/dev/boards/relations.py
geverghe: So what is the check and what does it show as error now?

gauravsaralMs: if any id is incorrect, we will show this
`'Id(s) supplied in --target-ids is not valid'`

azure-devops/azext_devops/dev/boards/relations.py
gauravsaralMs: well there are multiple reasons why this is should not be done.
1. say a WI has 20 relations (same type)
then the `Target ID` column will grow and given that they are distinct IDs and not same name so trimming that info is not correct
2. say a WI has 5 relations of 4 different type
How will be present this info in one column? 
Only possible way will be to show multiple columns for same work item, but that will give false impression to user that there are multiple work items

wheel is present in the PR build as artifact

If at all this seems like something which user wants to know I can suggest we show relations_count in the table

gauravsaralMs: ahh, got it now
we can do that

gauravsaralMs: now I got the requirement.
maybe this is a valid case.

but then should we also have a show for relations? because someone just want to see relation without add/remove?

atbagga: e.g. In build tag add or build tag list we only list down all tags and no other information. 
Here we can list artifacts and their relationship type. 

Where I am getting the wheel from.. 
```hub pr checkout 541```
And invoke from inside the path **azure-devops>** I do not even need to build the wheel :)

Comment (gauravsaralMs) [reply azure-devops/azext_devops/dev/boards/relations.py]: done, but still stuck with showing Url because that is what we get in response from service and given that it can be any Url, I am trying not to manipulate it

image

Comment (geverghe) [reply azure-devops/azext_devops/dev/boards/relations.py]: Fair enough for first try

File: azure-devops/azext_devops/dev/boards/arguments.py [hunk 8]
```diff
@@ -20,3 +20,14 @@ def load_work_arguments(self, _):
     with self.argument_context('boards work-item delete') as context:
         context.argument('yes', options_list=['--yes', '-y'], action='store_true',
                          help='Do not prompt for confirmation.')
+
+    with self.argument_context('boards work-item relation') as context:
+        context.argument('id', help='The ID of the work item')
+        context.argument('target_ids', help='ID(s) of work-items to create relation with. \
```
geverghe: Can you ensure the parameter is in the singular - target-id?

atbagga: But it can accept multiple values. We have this at other places as well. 
image

geverghe: 😟 - Singular is a parameter paradigm - can't locate the documentation location but I remember from a conversation with the Azure CLI team; we'll need to correct this as well.

azure-devops/azext_devops/dev/boards/commands.py
geverghe: Perfect - makes sense. Need to ensure there are no groups with single commands unless explicitly required.

azure-devops/azext_devops/dev/boards/arguments.py
```diff
@@ -20,3 +20,14 @@ def load_work_arguments(self, _):
...
+                         Multiple values can be passed comma seperated. Example: 1,2 ')
```
geverghe: Typo - should be sep**a**rated

geverghe:
I am not sure if I see all the documentation - here is a set of documentation that I have verified:
Manage work item relations.
id - The ID of the work item
target-id - ID(s) of work-items to create relation with. Multiple values can be passed comma separated. Example: 1,2 ')
List work item relations supported in the organization.

Am I missing something?

gauravsaralMs:
> I am not sure if I see all the documentation - here is a set of documentation that I have verified:
> Manage work item relations.
> id - The ID of the work item
> target-id - ID(s) of work-items to create relation with. Multiple values can be passed comma separated. Example: 1,2 ')
> List work item relations supported in the organization.
> 
> Am I missing something?

looks good

Issue #333, opened 2019-01-16
Pipelines commands - Create, update, list and run a pipeline; list, show and tag pipeline runs


atbagga:
|  Command | Description  	|   	
|---	|---	|	
| az pipelines create   	|   Create a pipeline (both from YAML & a GIT repo) 	|   
| az pipelines  delete 	|   Delete a pipeline |     	
| az pipelines update       |  Open the pipeline YAML in default editor. 	|   
| az pipelines run             |  Run a pipeline manually  	|   
| az pipelines list              |   List all pipelines and their status |   
| az pipelines show          |  Show details pertaining to a pipeline  	|   
| az pipelines runs list |  List of pipeline runs |
| az pipelines runs show |  Show details of a pipeline run |
| az pipelines runs tag add |  Add tags to a pipeline run |
| az pipelines runs tag list |  List tags assigned to a pipeline run|
| az pipelines runs tag delete |  Delete a tag from a pipeline run |

jaraco:
In [this document](https://marketplace.visualstudio.com/items?itemName=ms-vsts.cli), which links to this repo, it suggests that the `vsts build` command is used for working with pipelines... but the set above refers to a different command and subcommand. Where should I be hanging my hopes for a CLI for managing pipelines in dev.azure.com?

atbagga:
@jaraco  Firstly apologies for outdated documentation. We will have it updated soon. 

The [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest) with the [Azure DevOps extension](https://github.com/Microsoft/azure-devops-cli-extension) has replaced the VSTS CLI. The VSTS CLI has been deprecated and will no longer be receiving new features. We recommend that users of the VSTS CLI switch to the Azure CLI and add the Azure DevOps extension. To get started refer the quick start guide in the [readme](https://github.com/Microsoft/azure-devops-cli-extension/blob/master/README.md) of this repo. 

The commands which were supported in VSTS-CLI are all available in the Azure DevOps Extension.
So vsts build list is now az pipelines build list and so on. You can checkout the existing pipelines command [here](https://docs.microsoft.com/en-us/cli/azure/ext/azure-devops/pipelines?view=azure-cli-latest). 

This feature request is to enhance the pipeline commands to support yml based definitions too and letting you create and update pipelines which was not supported in VSTS CLI.   

Let us know if you face any issues in getting started.

jaraco:
[...]

PR #549, opened 2019-04-24
Pipeline create 

atbagga:

#333


azure-devops/azext_devops/dev/common/github_credential_manager.py
```diff
@@ -0,0 +1,79 @@
...
+class GithubCredentialManager():
+    """ GithubCredentialManager
+    """
+    def __init__(self):
+        self.username = None
+        self.password = None
+        self.token = None
+
+    def create_token(self, note=None):
+        self.username = prompt(msg='Enter your GitHub username (leave blank for using already generated PAT): ')
```
gauravsaralMs: when asking for PAT, should we also give expectation on the scopes required? users might not be very happy to give PAT will all scopes

azure-devops/azext_devops/dev/common/github_credential_manager.py
```diff
@@ -0,0 +1,79 @@
...
+
+logger = get_logger(__name__)
+
+
+@singleton
+class GithubCredentialManager():
+    """ GithubCredentialManager
+    """
+    def __init__(self):
+        self.username = None
```
gauravsaralMs: are these really needed as class level vars?

azure-devops/azext_devops/dev/common/github_credential_manager.py
```diff
@@ -0,0 +1,79 @@
...
+                'repo',
+                'user'
+                ],
+            "note": note
+            }
+        headers = {'Content-Type': 'application/json' + '; charset=utf-8',
+                   'Accept': 'application/json',
+                   'Authorization': basic_auth}
+        response = self.post_authorization_request(headers=headers, body=request_body)
+        if (response.status_code == 401 and response.headers.get('X-GitHub-OTP') and
```
gauravsaralMs: it will be good to add comments on these logics on where to get this logic from.
I think this is coming from 
https://developer.github.com/v3/auth/

azure-devops/azext_devops/dev/pipelines/github_api_helper.py
```diff
@@ -0,0 +1,146 @@
...
+def create_github_branch(repo, source):
+    token = get_github_pat_token()
+    # Validate new branch name is valid
+    branch_is_valid = False
+    while not branch_is_valid:
+        new_branch = prompt(msg='Enter new branch name to create: ')
+        if not get_github_branch(repo, new_branch):
+            branch_is_valid = True
+        else:
+            logger.warning('Not a valid branch name.')
```
gauravsaralMs: do we want to show this info to user? I think logger will not show up and we should use prompt..
**not sure**

azure-devops/azext_devops/dev/common/utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+def open_file(filepath):
+    import subprocess
+    import platform
+    import os
+    if platform.system() == 'Darwin':       # macOS
+        subprocess.call(('open', filepath))
+    elif platform.system() == 'Windows':    # Windows
+        # disable pylint warning since we run pylint on linux agent and startfile is a windows only function
+        os.startfile(filepath)  # pylint: disable=no-member
```
gauravsaralMs: just confirming that this will handle appselector scenario also? where windows is opening yml file extension for first time and does not have any default app associated with it?

also this function will open file and wait for it to close, or just open and return?

azure-devops/azext_devops/dev/common/github_credential_manager.py
atbagga: Good point!
Though I was printing the warning message with the scopes and name of PAT created.
Moving it to before asking the credentials.

azure-devops/azext_devops/dev/common/github_credential_manager.py
atbagga: Yes, this is a singleton class so if the credentials are needed again during the command run we reuse the token from the class var.

azure-devops/azext_devops/dev/pipelines/github_api_helper.py
atbagga: Warning shows on the console.

azure-devops/azext_devops/dev/common/utils.py
atbagga: It will handle app selector by opening the choose app dialog. 
Function will open the file and exit, it does not wait. I will add a header with this info.

PR #552, opened 2019-04-24
Wiki command fixes - project wiki create scenario

atbagga:
1. Project wiki creation should not require --name as it is defaulted to project name


atbagga:
> Just thinking out loud, what if let the object pass to service without name?
> the error message will contains something related to name but not exactly --name.. it will be sub-optimal experience but then that will hold true for other cases, like invalid chars in wiki name

Yes that is true I had tried it before adding the check, I think it makes sense to handle truly required parameters in the command logic for good error messages. 

image

PR #568, opened 2019-04-30
Bug fixes : Handle long list of service connections and  Github file checkin failure  in pipeline create command

atbagga:
Bug Fixes in az pipeline create command: 
1. Handle long list of service connections in pipeline create (filter the list to only connections create by the user or App connections
1. Handle github file checkin failures by raising appropriate error

### DIFF
--- a/azure-devops/azext_devops/dev/pipelines/pipeline_create_helpers/pipelines_resource_provider.py
+++ b/azure-devops/azext_devops/dev/pipelines/pipeline_create_helpers/pipelines_resource_provider.py
@@ -22,6 +22,8 @@ def get_github_service_endpoint(organization, project):
     Service endpoint connection name is asked as input from the user, if the environment is non interative
     name is set to default  AzureDevopsCliCreatePipelineFlow
     """
+    from azext_devops.dev.common.identities import get_current_identity, get_account_from_identity
+    authenticated_user_unique_id = get_account_from_identity(get_current_identity(organization))
     se_client = get_service_endpoint_client(organization)
     existing_service_endpoints = _get_service_endpoints(organization, project, 'github')
     service_endpoints_choice_list = ['Create new GitHub service connection']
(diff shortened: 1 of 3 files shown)


azure-devops/azext_devops/dev/pipelines/pipeline_create_helpers/pipelines_resource_provider.py
```diff
@@ -19,36 +19,36 @@ def get_github_service_endpoint(organization, project):
...
     se_client = get_service_endpoint_client(organization)
     existing_service_endpoints = _get_service_endpoints(organization, project, 'github')
     service_endpoints_choice_list = ['Create new GitHub service connection']
     github_service_endpoints = []
     choice = 0
     for endpoint in existing_service_endpoints:
-        if endpoint.authorization.scheme == 'InstallationToken':
-            service_endpoints_choice_list.append('{} {}'.format(endpoint.name, '(Recommended)'))
-        else:
+        if (endpoint.authorization.scheme == 'InstallationToken' or
```
gauravsaralMs: `get_service_endpoints` in the client takes scheme as a param for filtering... can we use that one?

atbagga: I think I cannot use that filter here.. 

**Reason:** I need all service connections created by me or the ones with auth scheme = installationtoken. 
I will still need to pass all auth schemes to get_service_endpoints since I need my service endpoints which may or may not be InstallationToken scheme. 

Good catch on the functionality being available though. I had missed that.

atbagga:
> The user will end up seeing a different service endpoint list in CLI and in UI..
> 
> If we doing this just to get around friendly_prompt_list issue
> then I will suggest the following
> 
> 1. Do a proper fix in friendly_prompt_lis
> 2. Build some kind of loop in friendly_prompt_lis
>    like end option can be (see more options) basically pagination kind of thing

We will have to fix the list and I will do that. Here the intention is to actually reduce the options also since we do not want the user to have 400 options to choose from. 
These are the service endpoints which will be most likely used. 

I have raised  an issue to fix the list. #571

PR #580, opened 2019-05-06
changes for deleting service-endpoint

gauravsaralMs:

Fixes: https://github.com/Azure/azure-devops-cli-extension/issues/609


atbagga:
Another functionality which will be good to expose is filtering the service connections by type when listing the service endpoints. 
Not sure if there is a way to list out all the types today which will be possible values of the filter in the list command.

gauravsaralMs:
> Another functionality which will be good to expose is filtering the service connections by type when listing the service endpoints.
> Not sure if there is a way to list out all the types today which will be possible values of the filter in the list command.

Agree..
we can look into this later

Issue #585, opened 2019-05-07
[Bug] Parameter issue in "az pipelines runs artifact list"


s-bauer:
Referring to https://github.com/Azure/azure-cli-extensions/issues/676

I'm trying to get all artifacts for a given buildId. I use the following command:

```
az pipelines runs artifact list --run-id [BuildId] --org "https://dev.azure.com/[MyOrg]" --project "[MyProject]"
```

however it throws the error, that you can see in the referenced issue. However I figured out there's probably a parameter mapping problem. If i switch "--project" and "--run-id" it works:

```
az pipelines runs artifact list --project [BuildId] --org "https://dev.azure.com/[MyOrg]" --run-id "[MyProject]"
```

Please fix this!!

gauravsaralMs:
@s-bauer : can you please share debug logs?
az pipelines runs artifact list --run-id [BuildId] --org "https://dev.azure.com/[MyOrg]" --project "[MyProject]" --debug

s-bauer:
@gauravsaralMs There u go: https://gist.github.com/s-bauer/485dfd250c61656d8b37e6c6db6effe2

gauravsaralMs:
Thanks for reporting this @s-bauer . We will fix this ASAP

atbagga:
Fix is checked in. It should be available in the next release 0.9.0 and later.

s-bauer:
@atbagga When will 0.9.0 be released? Any eta yet?

PR #587, opened 2019-05-08
Fix issue for artifacts list and add tests

atbagga:

 - [x] : This PR has a corresponding issue open in the Repository. #585 


azure-devops/azext_devops/dev/pipelines/runs_artifacts.py
```diff
@@ -34,7 +35,7 @@ def run_artifact_list(run_id, organization=None, project=None, detect=None):
     """
     organization, project = resolve_instance_and_project(detect=detect, organization=organization, project=project)
     client = get_build_client(organization)
-    artifacts = client.get_artifacts(run_id, project)
+    artifacts = client.get_artifacts(project=project, build_id=run_id)
```
gauravsaralMs: I think it will be good if use named params while calling functions everywhere in our code..
Thoughts? 
@ishitam8  @atbagga

atbagga: Yes, more importantly when referencing external module functions. 
If we want to always do it.. lets check if pylint has such a rule.

Issue #560, opened 2019-04-25
Why do we have devops configure when there is already a az configure


atbagga:

atbagga:
Issues in moving under az configure-
1. Use-git-aliases besides being a boolean setting in our config also adds some specific mapping in the .gitconfig file. I doubt az configure will let us add that code there, nor would such an option make any sense in Azure Cli at such top level config command. As far as I can think there is no other workaround for this either.
1. Other issue I see with our default settings moving to az configure command is the naming clash. If az configure allows a common term like "Project" or "organization" set for devops extension, if some other extension independently tries to use the keys name in overloaded manner we might end up having confusing behavior.

e.g. az dms has a concept of project
```az dms project : Manage projects for an instance of the Data Migration Service```

PR #588, opened 2019-05-08
Add support to declare commands in preview

gauravsaralMs:

Use of commit is present here:
Hides service-endpoint command and help text for security
https://github.com/gauravsaralMs/azure-devops-cli-extension/commit/e03e0de2a5dd8da2f4459195c16728943d0f60b4

This is how it will function (as is supposed to be used):

image
image

Also seems like this will not create problem in release, where azure cli checks for help text etc
https://github.com/Azure/azure-cli-extensions/pull/680


azure-devops/azext_devops/dev/common/config.py
```diff
@@ -42,6 +44,14 @@ def set_global_config_value(section, option, value):
     azdevops_config.config_parser.read(AZ_DEVOPS_GLOBAL_CONFIG_PATH)
 
 
+def get_preview_value():
```
atbagga: Should have caching since this will affect -h commands which people expect to be fast.

azure-devops/azext_devops/dev/team/arguments.py
```diff
@@ -62,6 +62,7 @@ def load_team_arguments(self, _):
                          help='Do not prompt for confirmation.')
     with self.argument_context('devops configure') as context:
         context.argument('use_git_aliases', **enum_choice_list(_YES_NO_SWITCH_VALUES))
+        context.argument('enable_preview_commands', **enum_choice_list(_YES_NO_SWITCH_VALUES))
```
atbagga: Should we have the three_state_flag instead of yes/no? 
PM call. @geverghe

atbagga:
Couple of points here - 
1. We should update some documentation in getting started to take note of this.
1. Lets close #560  with @geverghe before we take more dependency on 'az devops configure'

Issue #548, opened 2019-04-23
[Bug] Allow branch policies to be bypassed when completing a pull request


jackjfrost:
It appears that there is no way to bypass branch policies when completing a pull request via the az devops CLI, even if the pull request has the flags set that should allow the policies to be bypassed.
```
% az repos pr complete --id 2820
This pull request requires an explicit override to bypass policies that aren't currently passing.
```

When showing the attributes of that pull request, the following is returned:
```
  "completionOptions": {
    "bypassPolicy": true,
    "bypassReason": "force",
    "deleteSourceBranch": true,
    "mergeCommitMessage": "forcing",
    "mergeStrategy": "squash",
    "squashMerge": true,
    "transitionWorkItems": true
  },

```

Assuming the user attempting to bypass the policies is in a group that has those permissions (Bypass policies when completing pull requests) allowed for the target repo, I would think that the 'complete' option should succeed.

atbagga:
Fix is merged to master. Should be available in version 0.9.0 and later.
ETA for release Week of 17th May.

jackjfrost:
This appears to be broken again:

~$ az repos pr update --org <myorg> --id 13469 --status "completed" --bypass-policy true --bypass-policy-reason "force"
Core Promote Pipeline must succeed to update master.

jackjfrost:
I also tried doing the above in two steps and that didn't work either:

~$ az repos pr update --org <myorg> --id 13469 --bypass-policy true --bypass-policy-reason "force"
~$ az repos pr show --id 13469 --org <myorg> | grep bypass -i
    "bypassPolicy": true,
    "bypassReason": "force",
~$ az repos pr update --org <myorg> --status completed --id 13469
Core Promote Pipeline must succeed to update master.

jackjfrost:
Nevermind, found my issue.

PR #589, opened 2019-05-08
Pull request completion not working due to bypass policy 

atbagga:

 - [ ] : This PR has a corresponding issue open in the Repository.  #548 


gauravsaralMs:
> This was not working since the API requires us to send the information during the completion patch request.

So this is a regression on API side? I mean we are calling API with a certain version so that behavior should not change

atbagga:
> > This was not working since the API requires us to send the information during the completion patch request.
> 
> So this is a regression on API side? I mean we are calling API with a certain version so that behavior should not change

I haven't checked with 4.0 but it could be an old issue too OR it got introduced with 5.0 API change.

PR #594, opened 2019-05-11
Removed preview text for Universal Packages help.

mitchdenny:
I was using Universal Packages and noticed that the preview message is still in the help. Thought I would submit a PR to address that now that it is GA.


mitchdenny:
Ping @zarenner

PR #613, opened 2019-05-22
Standardize flags and three state usage

gauravsaralMs:

Changes:
All policy commands
devops extension list : change in list extension commands
devops user add : change in add user command
devops configure : Git alias

all detect params


azure-devops/azext_devops/dev/repos/arguments.py
```diff
@@ -35,9 +33,9 @@ def load_code_arguments(self, _):
                          'exact match on the --branch argument. ' +
                          'If value is \'prefix\' the policy is applied across all branch folders that ' +
                          'match the prefix provided by the --branch argument.')
-        context.argument('is_blocking', arg_type=get_enum_type(_TRUE_FALSE_SWITCH),
+        context.argument('is_blocking', arg_type=get_three_state_flag(),
```
atbagga: Are we not changing the param names is_blocking -- blocking?

azure-devops/azext_devops/dev/team/arguments.py
```diff
@@ -89,8 +87,7 @@ def load_team_arguments(self, _):
     with self.argument_context('devops user') as context:
         context.argument('license_type', arg_type=get_enum_type(_LICENSE_TYPES))
     with self.argument_context('devops user add') as context:
-        context.argument('send_email_invite', arg_type=get_enum_type(_TRUE_FALSE_SWITCH),
-                         help='Whether to send email invite for new user or not.')
+        context.argument('send_email_invite', help='Whether to send email invite for new user or not.')
```
atbagga: arg_type?

azure-devops/azext_devops/dev/team/arguments.py
```diff
@@ -129,10 +126,8 @@ def load_team_arguments(self, _):
                          for given user/group and token.')
 
     with self.argument_context('devops extension') as context:
-        context.argument('include_built_in', arg_type=get_enum_type(_TRUE_FALSE_SWITCH),
-                         help='Include built in extensions.')
-        context.argument('include_disabled', arg_type=get_enum_type(_TRUE_FALSE_SWITCH),
-                         help='Include disabled extensions.')
+        context.argument('include_built_in', help='Include built in extensions.')
+        context.argument('include_disabled', help='Include disabled extensions.')
```
atbagga: arg_type?

atbagga: Ok.. We are converting this from string to bool now, given the default  in  the function definition.

azure-devops/azext_devops/dev/team/user.py
```diff
@@ -84,7 +83,7 @@ def add_user_entitlement(email_id, license_type, send_email_invite='true', organ
     :rtype: UserEntitlementsPatchResponse
     """
     do_not_send_invite = False
```
atbagga: Don't need this initialization.

azure-devops/azext_devops/dev/repos/arguments.py
gauravsaralMs: is_blocking is member of PolicyConfiguration and a property, so I am not sure what is the reason behind changing it's name just to create more confusion for users

gauravsaralMs: cc: @snankani  @geverghe : let me know if we should change --is-blocking to --blocking

azure-devops/azext_devops/dev/team/arguments.py
gauravsaralMs: yup..
no action here

azure-devops/azext_devops/dev/team/arguments.py
gauravsaralMs: no action here

azure-devops/azext_devops/dev/team/user.py
gauravsaralMs: nice catch :).
Fixed

gauravsaralMs:
> Please change error message when organization is not specified and detect is off
> services.py [_team_organization_arg_error]
> 
> DetectOnMsg

really nice catch :)

azure-devops/azext_devops/dev/repos/arguments.py
geverghe: IMO it should change to --blocking provided this is adhering to the registering flag (https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#registering-flags) paradigm listed here. The flag is named as bool_prop_enabled so that it makes sense without having to provide the actual value of true which is assumed here.

With adherence to the registering flags paradigm, now users need not provide the default value. Therefore, the command should make sense with the absence of an explicit value. Therefore here:
1. <With rename> az repos policy --blocking
2. <without rename> az repos policy --is-blocking
In the first, blocking gives clarity about the action that is being taken. Is blocking is still ambiguous about the action being taken - is it blocking? Is not blocking?

snankani: --blocking is more precise compared to --is-blocking in this case.

gauravsaralMs: will update to --blocking and --enabled but just putting my concerns
1. This is already rolled out and being changed without any feedback.. so any existing script using this will be broken
2. If I have to use it in any script then I will have to deal with 2 things now 'blocking' and 'is-blocking' because 'is-blocking' is part of the response from service.

PR #621, opened 2019-05-24
add support for import from private source

gauravsaralMs:

Fixes: https://github.com/Azure/azure-devops-cli-extension/issues/617


azure-devops/azext_devops/dev/repos/import_request.py
```diff
@@ -4,33 +4,78 @@
...
+
+    delete_se_after_import = False
+
+    password = None
+    import random
+    import string
+    import os
+    if requires_authorization and git_service_endpoint_id is None:
+        delete_se_after_import = True
+        GIT_SOURCE_PASSWORD_OR_PAT = CLI_ENV_VARIABLE_PREFIX + 'GIT_SOURCE_PASSWORD_OR_PAT'
```
atbagga: Suggest: Put this in const so all env variables are at one place.

azure-devops/azext_devops/dev/repos/import_request.py
```diff
@@ -4,33 +4,78 @@
...
+    :type git_service_endpoint_id: str
     """
     organization, project, repository = resolve_instance_project_and_repo(
         detect=detect,
         organization=organization,
         project=project,
         repo=repository,
         repo_required=True)
+
+    delete_se_after_import = False
```
atbagga: Why not take this as user input?

azure-devops/azext_devops/test/repos/test_import_request.py
```diff
@@ -0,0 +1,81 @@
...
+                                            
+from azext_devops.dev.common.services import clear_connection_cache
+from azext_devops.test.utils.authentication import AuthenticatedTests
+from azext_devops.test.utils.helper import get_client_mock_helper
+
+
+class TestRepositoryMethods(AuthenticatedTests):
+
+    _TEST_DEVOPS_ORGANIZATION = 'https://someorg.visualstudio.com'
+    _TEST_PAT_TOKEN = 'lwghjbj67fghokrgxsytghg75nk2ssguljk7a78qpcg2ttygviyt'
```
atbagga: Assuming this is a bad token.

azure-devops/azext_devops/dev/repos/import_request.py
gauravsaralMs: it's like if we create we delete it, else we don't
very much like UI (where this is not exposed as well)

azure-devops/azext_devops/test/repos/test_import_request.py
gauravsaralMs: should be, but still changed

azure-devops/azext_devops/dev/repos/import_request.py
atbagga: Rest API gives you an option to delete after use, so I was comparing with that. But sounds good to me to hide that param. People can always script it using the delete service endpoint command.

azure-devops/azext_devops/test/repos/test_import_request.py
```diff
@@ -0,0 +1,81 @@
...
+from azext_devops.devops_sdk.v5_0.git.models import GitImportRequest
+
+from azext_devops.dev.repos.import_request import create_import_request
+                                            
+from azext_devops.dev.common.services import clear_connection_cache
+from azext_devops.test.utils.authentication import AuthenticatedTests
+from azext_devops.test.utils.helper import get_client_mock_helper
+
+
+class TestRepositoryMethods(AuthenticatedTests):
```
ishitam8: Test**Import**RepositoryMethods

gauravsaralMs: Fixed , nice catch

PR #625, opened 2019-05-24
Including missing scripts and fixing markdown linting

geverghe:
1. Fixing markdown linting issues
2. Including scripts that were missed which are required for the scaffolding script


examples/scaffolding.md
```diff
@@ -1,3 +1,5 @@
+# Automating Team Project set up
+
 **Scaffolding.ps1** is a powershell script which helps users of an organization to have a standard getting started experience.
```
atbagga: Will be good to mention the dependency scripts since they are not separated into some folder structure.

geverghe: Based on @ishitam8 recommendation, I have moved everything inside a single folder - Scaffolding with all the dependent scripts. Once the scripts are truly generalized, we can move it outside the folder as individual component scripts.

Issue #593, opened 2019-05-10
[Bug] git aliases don't work


centreboard:
**Describe the bug**
Trying to use git aliases results in 
> environment: line 0: exec: az: not found

**To Reproduce**
Azure Cli Version: 2.0.64
Azure-Devops extension version: 0.8.0

Steps to reproduce the behaviour:
```
> az devops configure --use-git-aliases yes
> git pr create
environment: line 0: exec: az: not found
```

**Expected behaviour**
`git pr` should have same behaviour as `az repos pr`
`git repo` should have same behaviour as `az repos`

**Additional info**
Reproduced on windows in cmd and pwsh.

from .gitconfig
```
[alias]
	pr = "!f() { exec az repos pr \"$@\"; }; f"
	repo = "!f() { exec az repos \"$@\"; }; f"
```

Jaykul:
The `az` command in the alias needs a `.cmd` added to it on Windows.

atbagga:
@centreboard @Jaykul  How did you install azure cli? Was it using the msi or just a pip install?

centreboard:
@atbagga I used the msi

qzdl:
The file is https://github.com/Azure/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/dev/common/git.py

Encountering this was a bit of a pain

gauravsaralMs:
@qzdl , @centreboard  is this happening when you are trying to use git from git bash?
Then this might be related to an issue in azure cli
https://github.com/Azure/azure-cli/issues/8560 

can you confirm if this happens from normal terminal also or just git bash..
Apologies for the questions but this will help us to root cause it correctly.

centreboard:
I've been running from powershell/cmd prompts. 
`az repos pr create` works from them.
`git pr create` only works once I've edited the git config to use `az.cmd`

gauravsaralMs:
looks like ultimately git alias is executed with git bash environment only
![image](https://user-images.githubusercontent.com/43203065/58274568-c8515400-7db0-11e9-89aa-696db1f3c255.png)

this proves that because cls is not recognized but clear works in cmd (which can't happen in cmd, but happens in bash)

atbagga:
This is related to Azure CLI MSI not packaging/handling the az bash script. It does work for me on my windows machine since az bash script is picked from the python scripts folder. 

One quick fix for this could be to add windows specific hack in setting git aliases as suggested above and use az.cmd instead of az. 

[...]

PR #626, opened 2019-05-24
Git alias fix for windows and tests

atbagga:

 - [x] : This PR has a corresponding issue open in the Repository. #593 

In our pipeline the windows machine does not use the MSI installer hence the issue is not reproducible there. 
Just to show a test fail scenario I changed the code to alias the command as az1 instead of az and here is the pipeline failure link - https://dev.azure.com/ms/azure-devops-cli-extension/_build/results?buildId=16747&view=ms.vss-test-web.build-test-results-tab

### DIFF
--- a/tests/test_gitAliasTest.py
+++ b/tests/test_gitAliasTest.py
@@ -1,0 +1,12 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+import subprocess
+from azure.cli.testsdk import ScenarioTest
+
+class TestGitAliasing(ScenarioTest):
+    def test_git_aliases_pr_repo(self):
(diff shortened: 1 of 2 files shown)


tests/test_gitAliasTest.py
```diff
@@ -0,0 +1,12 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+import subprocess
+from azure.cli.testsdk import ScenarioTest
+
+class TestGitAliasing(ScenarioTest):
+    def test_git_aliases_pr_repo(self):
+        self.cmd('az devops configure --use-git-aliases')
+        repo_help = subprocess.check_output('git repo -h', shell=True)
```
gauravsaralMs: What is the difference between `self.cmd` and `subprocess.check_output` ? can we just use self.cmd everywhere?

atbagga: self.cmd is implemented in ScenarioTest and it checks that the command to be executed is starting with 'az' :)

PR #640, opened 2019-05-28
Update clients to get bug fixes in python client

atbagga:
Mainly to get a fix for issue https://github.com/microsoft/azure-devops-python-api/issues/205
I had done a [manual fix ](https://github.com/Azure/azure-devops-cli-extension/commit/a1c79b157618c787f8bc1a9a529881ec8619473d) on our end to handle this for wiki client. The update takes care of handling it for all clients and for all additional header keys. 


atbagga:
@gauravsaralMs We should react to this in invoke command to allow passing additional headers that will allow a few commands which are blocked by invoke right now. I will take that in  a separate PR. 

@ishitam8  I have cherry picked your overwrite of one of the model files. Assuming that was the only change of manual overwrite. Please verify once.

Issue #581, opened 2019-05-06
[Bug] File encoding issue when CLI takes arguments via file. e.g. az devops wiki page create , az devops invoke 


atbagga:
**Describe the bug**
I cannot use utf-8 encoded file to create a wiki page

**To Reproduce**
Azure Cli Version: 2.0.63
Azure-Devops extension version: 0.6.0

1. Configure and login as usual
1. Create a file with content using powershell e.g.  echo "hello world" | Out-File -FilePath 'd:/a.txt'
1. Try to create a page with the file in #3 as content. 
e.g. az devops wiki page create --wiki myprojectwiki --path /mypage3 --file-path "d:/a.txt" 

Error - Page content looks like this - 
![image](https://user-images.githubusercontent.com/11888714/57219892-1b888000-7017-11e9-9041-f4130f60a3f6.png)

**Expected behavior**
The page should be created with correct content.

atbagga:
Possible solutions: 

1. Mention the encoding as utf-8 in the  command help text and read the file as utf-8 else error out appropriately.
1. Allow a -encoding parameter with values like utf-8, utf-16, ascii etc and read the file appropriately. User will have option to use any encoding in the list.
```az keyvault secret set -h``` uses option #2 to allow users to enter the encoding type. See help for reference.

Sample help text from 'az keyvault secret set -h'

``` 

Content Source Arguments
    --encoding -e           : Source file encoding. The value is saved as a tag (`file-
                              encoding=<val>`) and used during download to automatically encode the
                              resulting file.  Allowed values: ascii, base64, hex, utf-16be,
                              utf-16le, utf-8.  Default: utf-8.
    --file -f               : Source file for secret. Use in conjunction with '--encoding'.
    --value                 : Plain text secret value. Cannot be used with '--file' or '--encoding'.
```

SathishRanjan:
why cant we autodetect it? we do that in server side

atbagga:
I did look for options to auto-detect and it seems correctly detecting the encoding at all times is impossible. 

There is nothing in the python standard library, though there are some other third party modules to auto detect but not with 100% accuracy (e.g. chardet library) and not without performance cost.

atbagga:
@snankani @geverghe For triage. This got skipped in today's triage. 
Should we go with approach 2 suggested above?

geverghe:
[...]

PR #641, opened 2019-05-28
Adding encoding parameter to allow non ascii encoded files to be allowed

atbagga:

 - [x] : This PR has a corresponding issue open in the Repository. #581 

### DIFF
--- a/azure-devops/azext_devops/dev/common/utils.py
+++ b/azure-devops/azext_devops/dev/common/utils.py
@@ -18,20 +18,13 @@ def read_file_content(file_path, encoding):
     if encoding in FILE_ENCODING_TYPES:
         with open(file_path, 'r', encoding=encoding) as f:
             try:
-                content = f.read()
+                return f.read()
             except UnicodeDecodeError as ex:
                 logger.debug(msg=ex)
                 raise CLIError("Unable to decode file '{}' with '{}' encoding.".format(
                     file_path, encoding))
-            encoded_str = content
-            encoded = content.encode(encoding)
-            decoded = encoded.decode(encoding)
-
-    if content != decoded:
-        raise CLIError("Invalid encoding '{}'".format(encoding))
-
-    return encoded_str
-
+    else:
+        raise CLIError("File encoding {encoding} is not supported.".format(encoding=encoding))
 
 
 def open_file(filepath):
--- a/azure-devops/azext_devops/dev/common/utils.py
+++ b/azure-devops/azext_devops/dev/common/utils.py
@@ -19,7 +19,8 @@ def read_file_content(file_path, encoding):
         with open(file_path, 'r', encoding=encoding) as f:
             try:
                 content = f.read()
-            except UnicodeDecodeError:
+            except UnicodeDecodeError as ex:
+                logger.debug(msg=ex)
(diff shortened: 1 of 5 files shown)


atbagga:
@gauravsaralMs @ishitam8  If this looks good I will add the same for invoke command where we take input via file.

azure-devops/azext_devops/dev/common/utils.py
```diff
@@ -2,6 +2,35 @@
...
+    if not file_path or not encoding:
+        logger.debug("File path (%s) or encoding (%s) is missing.", file_path, encoding)
+        return None
+
+    if encoding in FILE_ENCODING_TYPES:
+        with open(file_path, 'r', encoding=encoding) as f:
+            try:
+                content = f.read()
+            except UnicodeDecodeError:
+                raise CLIError("Unable to decode file '{}' with '{}' encoding.".format(
```
gauravsaralMs: concern: we are eating the exception here.
maybe we can log it

azure-devops/azext_devops/dev/common/utils.py
```diff
@@ -2,6 +2,35 @@
...
+                    file_path, encoding))
+            encoded_str = content
+            encoded = content.encode(encoding)
+            decoded = encoded.decode(encoding)
+
+    if content != decoded:
```
gauravsaralMs: is this check always required? seems like an overkill..
may be we can give caller an option if this kind of check is required or not

azure-devops/azext_devops/dev/team/wiki.py
```diff
@@ -137,16 +139,15 @@ def add_page(wiki, path, comment=_DEFAULT_PAGE_ADD_MESSAGE, content=None, file_p
...
+    elif file_path:
+        from azext_devops.dev.common.utils import read_file_content
+        parameters.content = read_file_content(file_path=file_path, encoding=encoding)
     return wiki_client.create_or_update_page(parameters=parameters, wiki_identifier=wiki,
                                              project=project, path=path, version=None, comment=comment)
 
 
 def update_page(wiki, path, version, comment=_DEFAULT_PAGE_UPDATE_MESSAGE, content=None, file_path=None,
-                organization=None, project=None, detect=None):
+                encoding='utf-8', organization=None, project=None, detect=None):
```
gauravsaralMs: this will need some change in the example as well?

gauravsaralMs:
> @gauravsaralMs @ishitam8 If this looks good I will add the same for invoke command where we take input via file.

yup.. lets fix that as well

azure-devops/azext_devops/dev/common/utils.py
atbagga: We are re-raising custom error. Not eating the exception.

atbagga: Will add debug trace.

azure-devops/azext_devops/dev/team/wiki.py
atbagga: Not really.. since this has a default value. Current examples will continue to work with copy paste.

azure-devops/azext_devops/dev/common/utils.py
atbagga: Getting rid of it for now. We can revisit more strict checks if we see some issues.

Issue #655, opened 2019-06-03
[Bug] invoke command fails if the api response does not have a content-type header e.g. release definition delete


atbagga:
**Describe the bug**
invoke command fails if the api response does not have a content-type header e.g. release definition delete

**To Reproduce**

1. login using az login
2. setup defaults..
3. Run command....
```
az devops invoke --area release --resource definitions --route-parameters project=<PROJECT_NAME> --query-parameters definitionId=<DEFINITION_ID> --http-method delete
```
4. See error-

The command failed with an unexpected error. Here is the traceback:

argument of type 'NoneType' is not iterable
..
..
..
File "/Users/atulbagga/.azure/cliextensions/azure-devops/azext_devops/dev/team/invoke.py", line 112, in invoke
    if 'json' in response.headers.get("content-type") and not out_file:
TypeError: argument of type 'NoneType' is not iterable

**Expected behavior**
Command should succeed

atbagga:
Fixed in master. Will be available in next release 0.11.0.

PR #656, opened 2019-06-03
Bug fix:Invoke command fails if content-type is none in response.

atbagga:

 - [x] : This PR has a corresponding issue open in the Repository. #655 

### DIFF
--- a/azure-devops/azext_devops/dev/team/invoke.py
+++ b/azure-devops/azext_devops/dev/team/invoke.py
@@ -15,7 +15,7 @@ from azext_devops.devops_sdk.client import Client
 logger = get_logger(__name__)
 
 
-# pylint: disable=too-many-locals, too-many-statements, inconsistent-return-statements, protected-access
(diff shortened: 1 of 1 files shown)


azure-devops/azext_devops/dev/team/invoke.py
```diff
@@ -106,24 +106,29 @@ def invoke(area=None, resource=None,
...
+
+    if not response.headers.get("content-type"):
+        logger.info('Content type header is None.')
+        no_content = True
+    elif 'json' in response.headers.get("content-type") and not out_file:
         return response.json()
 
-    if not out_file:
-        raise CLIError('Response is not json, you need to provide --out-file where it can be written')
+    if not no_content:
```
snankani: Logic is correct. 

To be honest, took some time to review the combination of no_content and not. I felt it would be a bit easier if we can use something like isContentAvailable as it will make is easy to read and comprehend in review. 

I mean, instead of maintaining is content not available in response, we can use the variable to contain the value to represent if content is available. Just the other way.

atbagga: Done. Thanks!

Issue #688, opened 2019-06-21
Work Item details in table format is not showing any data in the 'Assigned To' field.


suhaib16:
**Description**
Work Item details in table format is not showing any data in the 'Assigned To' field.

**To Reproduce**
Azure CLI Version: 
Use az --version 
Look for something like-
azure-cli (2.0.64)

Azure-DevOps extension version: 
Use az --version
Look for something like-
Extensions:
azure-devops (0.8.0)

1. Login using az login
2. Setup defaults.
3. Run command:
`az boards query --wiql "select [System.Id], [System.Title], [System.AreaPath], [System.AssignedTo], [System.State] from workitems where [System.TeamProject] = '<project name>'" --org <organization URL>  -o table`
4. No data is being provided in the 'Assigned To' column.

**Screenshot**
![Capture](https://user-images.githubusercontent.com/25266173/59922935-ab5f7d80-944f-11e9-8118-542296dc4ba4.PNG)

atbagga:
Fix in 0.12.0. Release ETA Next week.

PR #691, opened 2019-06-24
Fix table output for workitem query output

atbagga:

 - [x] : This PR has a corresponding issue open in the Repository. #688 


azure-devops/azext_devops/dev/boards/_format.py
```diff
@@ -106,6 +106,8 @@ def transform_work_item_query_result_row_output(row):
                     if len(title) > _WORK_ITEM_TITLE_TRUNCATION_LENGTH:
                         title = title[0:_WORK_ITEM_TITLE_TRUNCATION_LENGTH - 3] + '...'
                     table_row[field_reference.name] = title
+                if field_reference.reference_name == 'System.AssignedTo':
```
snankani: should n't this be elif (else if) check?
if reference_name is System.Title then first title will be stored, then System.AssignedTo check will fail and value of title stored for field_reference.name will be overwritten with field_reference.reference_name.

atbagga: Nice Catch!

Issue #692, opened 2019-06-24
 PAT env variable  name is wrong in Samples.md 


atbagga:
1. PAT is picked from AZURE_DEVOPS_EXT_PAT not AZURE_DEVOPS_CLI_PAT
1. Step below can be removed from sample yaml if we just add Environment variable in the script running the actual command
 ```
- script: echo ${AZURE_DEVOPS_CLI_PAT} | az devops login
  env:
    AZURE_DEVOPS_CLI_PAT: $(System.AccessToken)
  displayName: 'Login Azure DevOps Extension'
```

Change the step-
```
- script: |
    az pipelines build list
    git pr list
  displayName: 'Show build list and PRs'
```

To 

```
- script: |
    az pipelines build list
    git pr list
  displayName: 'Show build list and PRs'
  env:
    AZURE_DEVOPS_EXT_PAT: $(System.AccessToken)
```

atbagga:
#518 is filed for point (2)

Issue #518, opened 2019-04-01
Fix samples.md to remove login step - since automation should be via PAT in environment variable 


atbagga:

atbagga:
This will also bypass issue #486.

PR #694, opened 2019-06-25
Fix Samples.md with correct usage of PAT in Azure Pipelines

geverghe:

Reference Issues - #692 , #518

### DIFF
--- a/doc/samples.md
+++ b/doc/samples.md
@@ -164,9 +164,9 @@ steps:
 - script: az extension add -n azure-devops
   displayName: 'Install Azure DevOps Extension'
 
-- script: echo ${AZURE_DEVOPS_CLI_PAT} | az devops login
+- script: echo ${AZURE_DEVOPS_EXT_PAT} | az devops login
   env:
-    AZURE_DEVOPS_CLI_PAT: $(System.AccessToken)
+    AZURE_DEVOPS_EXT_PAT: $(System.AccessToken)
   displayName: 'Login Azure DevOps Extension'
 
 - script: az devops configure --defaults organization=$(System.TeamFoundationCollectionUri) project=$(System.TeamProject) --use-git-aliases yes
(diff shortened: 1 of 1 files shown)


doc/samples.md
```diff
@@ -164,18 +164,16 @@ steps:
 - script: az extension add -n azure-devops
   displayName: 'Install Azure DevOps Extension'
 
-- script: echo ${AZURE_DEVOPS_CLI_PAT} | az devops login
-  env:
-    AZURE_DEVOPS_CLI_PAT: $(System.AccessToken)
-  displayName: 'Login Azure DevOps Extension'
 
-- script: az devops configure --defaults organization=$(System.TeamFoundationCollectionUri) project=$(System.TeamProject) --use-git-aliases yes
+- script: az devops configure --defaults organization=$(System.TeamFoundationCollectionUri) project=$(System.TeamProject) --use-git-aliases
```
atbagga: Couple of comments: 
1. Configuring project and organization is redundant since the script will run from inside a checked out git repo so project and organization will be picked from there. 
1. Should we mention the use of $(PAT) OR fix permissions of the build account if (System.AccessToken) does not have permission to do something.

geverghe: 1. But there is no checking out here in script right? Or do we need to add that as well to provide complete context? 
2. Considering it is an issue (it is an issue right?), we can include that documentation in the troubleshooting guide? Does that make sense? I mean I felt the theme/main value prop here is it help people understand how to use it from YAML.

atbagga: For 1. Git checkout happens in pipeline no matter what.. 
image

See list of steps where my yaml contains only a powershell task. Rest of the Tasks are implicit.

geverghe: Capturing notes from our discussion:
1. Will knock off configuration
2. Will include a section in using PAT document to indicate how PATs can be used in pipelines

geverghe: Fixed 1). Will be fixing 2 in official docs

PR #714, opened 2019-07-15
Update/ break service connection/endpoint create command

gauravsaralMs:

This is also adding capability to create azure rm connection using certificates.
Will update the UTs once this overall approach is finalized.


azure-devops/azext_devops/dev/team/service_endpoint.py
```diff
@@ -92,52 +84,86 @@ def create_service_endpoint(service_endpoint_type, authorization_scheme, name,
...
+    :type name: str
+    :param service_endpoint_configuration: Configuration file with service endpoint request.
+    :type authorization_scheme: str
+    :rtype: :class:`ServiceEndpoint <service_endpoint.v4_1.models.ServiceEndpoint>`
+    """
+    organization, project = resolve_instance_and_project(detect=detect,
+                                                         organization=organization,
+                                                         project=project)
+    client = get_service_endpoint_client(organization)
+    with open(service_endpoint_configuration) as f:
```
atbagga: File encoding? This is will cause issue with utf8 vs ascii encoded files.
We fixed it in https://github.com/Azure/azure-devops-cli-extension/pull/641 for other commands.

azure-devops/azext_devops/dev/team/service_endpoint.py
```diff
@@ -92,52 +84,86 @@ def create_service_endpoint(service_endpoint_type, authorization_scheme, name,
...
+    service_endpoint_authorization = EndpointAuthorization(
+        parameters={'accessToken': github_access_token},
+        scheme=SERVICE_ENDPOINT_AUTHORIZATION_PERSONAL_ACCESS_TOKEN)
+    service_endpoint_to_create = ServiceEndpoint(
+        authorization=service_endpoint_authorization,
+        name=name, type=SERVICE_ENDPOINT_TYPE_GITHUB, url=github_url)
+    return client.create_service_endpoint(service_endpoint_to_create, project)
+
+
+def create_service_endpoint(service_endpoint_configuration, organization=None,
```
atbagga: This will need some help doc or sample on how to form the configuration file..

azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -28,6 +28,16 @@ def load_team_help():
     short-summary: Manage service endpoints/service connections
     """
 
+    helps['devops service-endpoint azurerm'] = """
+    type: group
+    short-summary: Manage Azure RM service endpoints/service connections
+    """
+
+    helps['devops service-endpoint github'] = """
```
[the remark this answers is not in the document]
gauravsaralMs: yes, but we do anticipate to author "update" command for every "create" command at some point of time

azure-devops/azext_devops/dev/team/service_endpoint.py
```diff
@@ -92,52 +84,86 @@ def create_service_endpoint(service_endpoint_type, authorization_scheme, name,
...
-            'environment': 'AzureCloud',
-            'creationMode': 'Manual'
-        }
-        service_endpoint_to_create = ServiceEndpoint(
-            authorization=service_endpoint_authorization, data=service_endpoint_data,
-            name=name, type=SERVICE_ENDPOINT_TYPE_AZURE_RM, url='https://management.azure.com/')
+        service_endpoint_authorization.parameters['authenticationType'] = 'spnKey'
+        service_endpoint_authorization.parameters['serviceprincipalkey'] = azure_rm_service_principal_key
+    else:
+        with open(azure_rm_service_principal_certificate_path, "r") as f:
```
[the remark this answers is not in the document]
gauravsaralMs: tested with utf-8 and ansi.. let see if we can break this and then we can add option to pass encoding..  not adding it right now because if not needed than that will be unnecessary complication

azure-devops/azext_devops/dev/team/service_endpoint.py
```diff
@@ -92,52 +84,86 @@ def create_service_endpoint(service_endpoint_type, authorization_scheme, name,
...
+    organization, project = resolve_instance_and_project(detect=detect,
+                                                         organization=organization,
+                                                         project=project)
+    client = get_service_endpoint_client(organization)
+    if AZ_DEVOPS_GITHUB_PAT_ENVKEY not in os.environ:
+        error_message = 'Please pass GitHub access token in ' + AZ_DEVOPS_GITHUB_PAT_ENVKEY +\
+                        ' environment variable in non-interactive mode.'
+        verify_is_a_tty_or_raise_error(error_message)
+        github_access_token = prompt_pass('GitHub access token:', confirm=True)
+    else:
```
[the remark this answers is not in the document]
gauravsaralMs: done... nice catch

azure-devops/azext_devops/dev/team/service_endpoint.py
gauravsaralMs: agree, will work on this with @geverghe

azure-devops/azext_devops/dev/team/service_endpoint.py
gauravsaralMs: added comment that only utf-8 is supported. not a complexity which is required

azure-devops/azext_devops/dev/common/utils.py
```diff
@@ -10,22 +10,36 @@
 FILE_ENCODING_TYPES = ['ascii', 'utf-16be', 'utf-16le', 'utf-8']
 
 
+def read_file_content_ver3(file_path, encoding):
```
atbagga: make these private functions.

azure-devops/azext_devops/dev/team/_help.py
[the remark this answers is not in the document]
atbagga: I will let @geverghe decide.

azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -28,6 +28,16 @@ def load_team_help():
     short-summary: Manage service endpoints/service connections
     """
 
+    helps['devops service-endpoint azurerm'] = """
+    type: group
+    short-summary: Manage Azure RM service endpoints/service connections
+    """
```
atbagga: nit: service endpoints/connections 
For all help texts.. confirm with @geverghe.

geverghe: @atbagga - thanks for the catch. And yes, since we are breaking away into new commands can we stick with the product nomenclature rather than the API nomenclature? I do not have any data points/feedback for this, but I personally feel better about sticking with the product nomenclature since we remain consistent. @gauravsaralMs - let me know your thoughts.

azure-devops/azext_devops/dev/team/_help.py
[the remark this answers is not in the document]
geverghe: Proposal 1: az devops service-endpoint github create
Pros
a. All though only one command in group, there is an update API and leaves room for the update command if and when we decide to build it
Cons
b. One level deeper and not adhering to the guidance pertaining to if one command don't have a group
 
Proposal 2: az devops service-endpoint create-github-connection
Pros
a. Adherence to guidance and one level higher discover
Cons:
a. Need to deprecate this command when creating group
b. Maintenance cost
Would need to provide the command with deprecation warning for potentially one version (say 10)/reasonable amount of time, and hide the command on the release post that (11) and then wait till traffic dies down/reasonable amount of time before removing the command completely. During the time, in case there are any changes, there is maintenance cost as both - deprecated and new command, would need to be updated
 
Due to the extra costs associated with proposal 2 for limited gains - hypothesis at this step being the extra level of command depth does not affect user experience significantly, let us go with Proposal 1 and leave space for update. Ideally we should do update and thereby eliminate the conundrum and make service connection commands complete.

azure-devops/azext_devops/dev/team/_help.py
gauravsaralMs: If at all we have to point users to read response/request docs then they will read service endpoint 
also this change will flow across to other commands/code which will be a significant change.. 
Let me know what will be better for customers and we can do that

atbagga: Ohh.. My original comment was only meant for a minor text change.  **service endpoint/ service connection**  should be **service endpoint/connection.** :)

I agree with Gaurav on not trying to change the name since the API docs will always call it service-endpoint anyways. Just because UI label was renamed we will end up maintaining a deprecation cycle for a set of commands doesn't feel right.

doc/service_endpoint.md
```diff
@@ -0,0 +1,85 @@
...
+  "readersGroup": null,
+  "groupScopeId": null,
+  "serviceEndpointProjectReferences": null,
+  "operationStatus": null
+}
+```
+
+This request body should be saved in a file and that file can act as a template for creation of service endpoints of type "Docker Registry Service Connection"
+
+This file (after updating appropriate values like Name, ID or password) can be passed to `--azure-rm-service-principal-certificate-path` environment variable.
```
atbagga: unclear

geverghe: Can you include this document directly to microsoft docs - vsts-docs-pr repo? Let me know if you need help with this.

doc/service_endpoint.md
```diff
@@ -0,0 +1,85 @@
...
+
+If AAD application uses [certificate for authentification](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-certificate-credentials) then create .pem for the certificate and pass path to .pem file in `--azure-rm-service-principal-certificate-path` argument.
+
+.pem file can be created using openssl 
+
+```bash
+openssl pkcs12 -in file.pfx -out file.pem -nodes -password pass:<password_here>
+```
+
+## Create service endpoint using configuration point
```
geverghe: Nit: Can you start with this and then get to create github and azure rm service connection? General -> specific?

doc/service_endpoint.md
geverghe: Isn't this file supposed to be provided in --service-endpoint-configuration ?

gauravsaralMs: yes, changed

doc/service_endpoint.md
gauravsaralMs: as discussed offline, not changing

doc/service_endpoint.md
gauravsaralMs: moved

azure-devops/azext_devops/dev/team/_help.py
geverghe: Pretty late to this party - but my reasoning as below - @atbagga , @gauravsaralMs :
1. Reference the User Experience - UI, when referencing to labeling as fundamentally terminology is the CLI experience and therefore it has to be consistent across interaction mediums. In case of conflicts with REST API, I think it would be better to adhere to the UI over the REST API as UI is the predominant interaction medium. 
2. My initial thought was with the breaking of the command, the usage pattern has changed; therefore, you are breaking the commanding pattern at this point, then include the name change as well. The deprecation pattern of hiding the command in the current version preventing new intake but supporting the existing users ensures that we are compliant. In terms of maintenance, if the existing users require any additional features, experience, or fixes, we point them to the new command. Feel that this is an acceptable experience.

Since this release has gone, I guess it is a wait till some conflict or feedback happens.

PR #724, opened 2019-07-22
Fix flaky tests

ishitam8:


tests/test_devopsSecurityGroupTest.py
```diff
@@ -109,6 +109,7 @@ def test_devops_group_and_member_tests(self):
             # remove membership
             remove_membership = self.cmd('az devops security group membership remove --group-id '+ project_group_descriptor +' --member-id '+ project_group_descriptor3 +' -y -o json --detect false')
             
+            time.sleep(5)
```
gauravsaralMs: might consider putting polling instead of blind sleep, that way it can be tolerant to a larger extent but not every time

ishitam8: As discussed with you, I shall add the polling instead of blind sleep for commands which rarely fail [Like boards, iterations]. Doesn't make much sense in updating for groups, since these APIs are usually slow [which is evident in UI as well].

PR #741, opened 2019-08-01
Pipeline create command should have an option to skip the first run

### LINKED ISSUE #740: [Feature Request] Skip first run of pipelines post creation
**Is your feature request related to a problem? Please describe.**
The sole purpose of this Azure Pipeline is to validate a PR. This is part of a process to setup a DevOps environment to new developer in the team. It creates an Azure repo, a feature branch (in addition to the master branch) and protect the master branch with the Azure Pipeline to validate the PR.
 
The problem is that immediately after creating the Pipeline, the pipeline is triggered (BTW it is marked as manually triggered). This is not at all a desired behavior, and there is no logic in triggering the pipeline while it is just being defined. In the case of the PR validation for instance, it may happen that it just fails because there is nothing to validate the PR pipeline is being defined.
 
I would like to know if there are plans to reconsider this behavior, this is considered as really problematic by the customer.

**Describe the solution you'd like**
Evaluate the possibility of providing a –skip-run option in az pipeline create command which a user can use to  not trigger the first run.

**Additional context**
I will propose a two fold answers, one about the incoherency of triggering the YAML pipeline immediately after the definition and the other one about my scenario.
 
Incoherency:
-	From the UI, the classic pipeline when defined is not triggered immediately after being saved. I suppose the arguments to go for this behavior in the classic pipeline are valid as well for the YAML pipelines… At minimum this is not coherent behavior.
-	After the YAML pipeline is saved the first time, it’s then possible to amend it and to save it without triggering it… If we take the logic below it should be triggered as well because in the same way after amending the pipeline it should require validation, authorizing resources,… This is another incoherent behavior.
-	When I create an incorrect YAML definition file for a pipeline (directly in the repo), and then create the pipeline using “az pipelines create” then it will create a kind of orphan pipeline object that is not usable (I could not find anything to do with it other than deleting it). That is a weird behavior, it just does not make sense.

My customer scenario:
-	My customer created a kind of pipeline generator repo with a master branch to which is attached a build definition. When a developer wants to deploy a micro-service project, he will create a feature branch in this pipeline generator repo, modify a file (deployment.info file) and raise a PR. A reviewer will review different things and validate the PR. Upon successful merge in the master of this pipeline generator repo, the build “linked” to the master will create a YAML Azure pipeline (for build and release purpose) in the target repo (the one in which the micro-service is being created) and will setup a secondary YAML Azure pipeline in the target repo for PR  validation purpose. The whole idea is that one YAML Azure pipeline will be used for gate quality purpose (the PR will succeed only if the quality gate is passed, and the other one will be used to build and release. Obviously, the last thing we want is to deploy when the quality gate is not passed… By having the YAML pipeline triggered immediately after saving the first definition it just breaks the whole logic.

atbagga:

 - [x] : This PR has a corresponding issue open in the Repository. Fixes #740 

### DIFF
--- a/azure-devops/azext_devops/dev/pipelines/_format.py
+++ b/azure-devops/azext_devops/dev/pipelines/_format.py
@@ -208,6 +208,17 @@ def transform_pipeline_runs_table_output(result):
     return table_output
 
 
+def transform_pipeline_or_run_table_output(result):
+    table_output = [_transform_pipeline_or_run_row(result)]
+    return table_output
+
+
+def _transform_pipeline_or_run_row(row):
+    if row.get("buildNumber", None):
+        return _transform_pipeline_run_row(row)
+    return _transform_pipeline_row(row)
+
+
 def transform_pipeline_run_table_output(result):
     table_output = [_transform_pipeline_run_row(result)]
     return table_output
--- a/azure-devops/azext_devops/dev/pipelines/_format.py
+++ b/azure-devops/azext_devops/dev/pipelines/_format.py
@@ -214,7 +214,7 @@ def transform_pipeline_or_run_table_output(result):
 
 
 def _transform_pipeline_or_run_row(row):
-    if row.get("buildNumber", None):
+    if row.get("buildNumber", None):  # Hack to detect the json object is definition or run
         return _transform_pipeline_run_row(row)
     return _transform_pipeline_row(row)
 
--- a/azure-devops/azext_devops/dev/pipelines/pipeline_create.py
+++ b/azure-devops/azext_devops/dev/pipelines/pipeline_create.py
@@ -46,7 +46,8 @@ _AZURE_GIT_REPO_TYPE = 'tfsgit'
 
 
 def pipeline_create(name, description=None, repository=None, branch=None, yml_path=None, repository_type=None,
-                    service_connection=None, organization=None, project=None, detect=None, queue_id=None):
+                    service_connection=None, organization=None, project=None, detect=None, queue_id=None,
+                    skip_first_run=None):
     """ (PREVIEW) Create a new Azure Pipeline (YAML based)
     :param name: Name of the new pipeline
     :type name: str
@@ -71,6 +72,9 @@ def pipeline_create(name, description=None, repository=None, branch=None, yml_pa
     :type service_connection: str
     :param queue_id: Id of the queue in the available agent pools. Will be auto detected if not specified.
     :type queue_id: str
+    :param skip_first_run: Specify this flag to prevent the first run being triggered by the command.
+    Command will return a pipeline if run is skipped else it will output a pipeline run.
+    :type skip_first_run: bool
     """
     repository_name = None
     if repository:
@@ -140,8 +144,10 @@ def pipeline_create(name, description=None, repository=None, branch=None, yml_pa
     created_definition = client.create_definition(definition=definition, project=project)
     logger.warning('Successfully created a pipeline with Name: %s, Id: %s.',
                    created_definition.name, created_definition.id)
-    return client.queue_build(
-        build=Build(definition=created_definition, source_branch=queue_branch), project=project)
+    if skip_first_run:
+        return created_definition
+    return client.queue_build(build=Build(definition=created_definition, source_branch=queue_branch),
+                                          project=project)
 
 
 def pipeline_update(name=None, id=None, description=None, new_name=None,  # pylint: disable=redefined-builtin
--- a/azure-devops/azext_devops/dev/pipelines/pipeline_create.py
+++ b/azure-devops/azext_devops/dev/pipelines/pipeline_create.py
@@ -45,6 +45,7 @@ _GITHUB_REPO_TYPE = 'github'
 _AZURE_GIT_REPO_TYPE = 'tfsgit'
 
 
+# pylint: disable=too-many-statements
 def pipeline_create(name, description=None, repository=None, branch=None, yml_path=None, repository_type=None,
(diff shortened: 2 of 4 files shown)


azure-devops/azext_devops/dev/pipelines/pipeline_create.py
```diff
@@ -71,6 +72,9 @@ def pipeline_create(name, description=None, repository=None, branch=None, yml_pa
     :type service_connection: str
     :param queue_id: Id of the queue in the available agent pools. Will be auto detected if not specified.
     :type queue_id: str
+    :param skip_first_run: Specify this flag to prevent the first run being triggered by the command.
+    Command will return a pipeline if run is skipped else it will output a pipeline run.
```
gauravsaralMs: `Command will return a pipeline if run is skipped else it will output a pipeline run.`
Is this something we are ok with? same command returning different type of response (based on different inputs)

azure-devops/azext_devops/dev/pipelines/_format.py
```diff
@@ -208,6 +208,17 @@ def transform_pipeline_runs_table_output(result):
     return table_output
 
 
+def transform_pipeline_or_run_table_output(result):
+    table_output = [_transform_pipeline_or_run_row(result)]
+    return table_output
+
+
+def _transform_pipeline_or_run_row(row):
+    if row.get("buildNumber", None):
```
gauravsaralMs: `if row.get("buildNumber", None):`
looks like a hack to detect response type based on presence of a field..
lets add a comment or can we come up with some better way to check type?

azure-devops/azext_devops/dev/pipelines/pipeline_create.py
```diff
@@ -140,8 +144,10 @@ def pipeline_create(name, description=None, repository=None, branch=None, yml_pa
     created_definition = client.create_definition(definition=definition, project=project)
     logger.warning('Successfully created a pipeline with Name: %s, Id: %s.',
                    created_definition.name, created_definition.id)
-    return client.queue_build(
-        build=Build(definition=created_definition, source_branch=queue_branch), project=project)
+    if skip_first_run:
```
gauravsaralMs: In general do this command really need to trigger the pipeline created?
Customers can always trigger the created pipeline after creating pipeline using this command.
The current customer feedback looks like a good enough indication to me as to why commands should be atomic in nature.
tldr: I would prefer if this command never queues the created pipeline.

azure-devops/azext_devops/dev/pipelines/pipeline_create.py
```diff
@@ -71,6 +72,9 @@ def pipeline_create(name, description=None, repository=None, branch=None, yml_pa
     :type service_connection: str
     :param queue_id: Id of the queue in the available agent pools. Will be auto detected if not specified.
     :type queue_id: str
+    :param skip_first_run: Specify this flag to prevent the first run being triggered by the command.
```
gauravsaralMs: will suggest default to be false, or have `queue_first_run` as a param.
The command is to create pipeline so doing anything above that (like queue it) should be optional (opt-in not opt-out)

azure-devops/azext_devops/dev/pipelines/pipeline_create.py
ishitam8: I agree, user can always queue the pipeline later on by running separate command.

azure-devops/azext_devops/dev/pipelines/pipeline_create.py
atbagga: So I see three concerns: 
1. Deterministic output
2. Command should be atomic and never trigger a run
3. If we have an option to trigger a run it should be opt-in rather than opt-out

For 1, we have had a thread with Travis Prescott on email which says - It is ok for a command to call different APIs and return different outputs. "From the perspective of the output, the output is still deterministic based on what the user specifies, so that is not a concern."
Is it not same as ```az devops service-endpoint show```?

For 2 and 3, the selected default behavior is due to two reasons:
1. Match UI behaviour since UI does not even give an option to opt-out, same is by default in the command.
2. Yaml issues are only found during the run so an invalid pipeline creation is caught early by triggering a run (which in most cases should be harmless if you are setting up a CI). It will catch issues like resource permissions etc.

@geverghe  To take a call. 

cc: @gauravsaralMs @ishitam8 @snankani

azure-devops/azext_devops/dev/pipelines/pipeline_create.py
atbagga: Resolving since I have captured it together under one comment.

azure-devops/azext_devops/dev/pipelines/pipeline_create.py
atbagga: Resolving since I have captured all feedback under one comment.

azure-devops/azext_devops/dev/pipelines/pipeline_create.py
geverghe: @atbagga has captured the reasoning quite accurately and I concur with the same. Few additional clarifications on point 2,3
1. I had a word with @vijayma, and the 80% scenario is that the experience for creating a pipeline is to have the pipeline built and the first run kick started. Without it, it's just a commit of another file. Further, the objective is to ensure at least one run to ensure things are in order.  
2. For the experience that is visualized, if we make it atomic, then the cognitive load to achieve the ideal experience increases from CLI. Considering we have a command working pretty efficiently at this point, I feel that we should be good tweaking the experience to include the skip part.

PR #773, opened 2019-09-02
Update PRReportGeneration.ps1

digitalinfinity:
The datetime parsing logic interprets the month as minutes instead and so the parsed datetime is incorrect causing the loop to exit prematurely. Small fix to interpret as month instead.


atbagga:
@digitalinfinity Thank you for the contribution.

Issue #781, opened 2019-09-10
[Feature Request] Add ability to get/add/delete pipeline folders, and list pipelines by folder


tylercranston:
**Is your feature request related to a problem? Please describe.**
Pipeline folders are useful for grouping permissions to pipelines. Currently to get/add/delete a pipeline folder you need to use REST, it would be useful to do this in CLI.

**Describe the solution you'd like**
e.g. If you want a new command to be added, what would it do?
Azure CLI commands to get/add/delete pipeline folders, and list pipelines by folder.

**Additional context**

PR #783, opened 2019-09-10
Pipeline folder management commands

atbagga:

 - [ ] : This PR has a corresponding issue open in the Repository. Fixes: #781 

### DIFF
--- a/azure-devops/azext_devops/dev/pipelines/_format.py
+++ b/azure-devops/azext_devops/dev/pipelines/_format.py
@@ -374,3 +374,25 @@ def _transform_pipeline_var_group_variable_row(key, value):
         val = val[0:_VALUE_TRUNCATION_LENGTH] + '...'
     table_row['Value'] = val
     return table_row
+
+
+def transform_pipelines_folders_table_output(result):
+    table_output = []
+    for item in result:
+        table_output.append(_transform_pipeline_folder_row(item))
+    return table_output
+
+
+def transform_pipelines_folder_table_output(result):
+    table_output = [_transform_pipeline_folder_row(result)]
+    return table_output
+
+
+def _transform_pipeline_folder_row(row):
+    table_row = OrderedDict()
+    table_row['Path'] = row['path']
+    val = row['description'] if row['description'] is not None else ''
+    if len(val) > _VALUE_TRUNCATION_LENGTH:
+        val = val[0:_VALUE_TRUNCATION_LENGTH] + '...'
+    table_row['Description'] = val
+    return table_row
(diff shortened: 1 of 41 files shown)


azure-devops/azext_devops/dev/pipelines/_format.py
```diff
@@ -374,3 +374,25 @@ def _transform_pipeline_var_group_variable_row(key, value):
...
+    table_output = [_transform_pipeline_folder_row(result)]
+    return table_output
+
+
+def _transform_pipeline_folder_row(row):
+    table_row = OrderedDict()
+    table_row['Path'] = row['path']
+    val = row['description'] if row['description'] is not None else ''
+    if len(val) > _VALUE_TRUNCATION_LENGTH:
+        val = val[0:_VALUE_TRUNCATION_LENGTH] + '...'
```
gauravsaralMs: should this be other way round?  show things from end and skip start because most probably multiple folders will share start..
or show some start and some end
start...end

azure-devops/azext_devops/dev/pipelines/pipeline_folders.py
```diff
@@ -0,0 +1,101 @@
...
+    :type project: str
+    :param detect: Automatically detect organization and project. Default is "on".
+    :type detect: str
+    """
+    organization, project = resolve_instance_and_project(
+        detect=detect, organization=organization, project=project)
+    client = get_build_client(organization)
+    from azext_devops.devops_sdk.v5_0.build.models import Folder
+    folder = Folder()
+    if description:
```
gauravsaralMs: is this `if` check required?

azure-devops/azext_devops/dev/pipelines/pipeline_folders.py
```diff
@@ -0,0 +1,101 @@
...
+    :param project: Name or ID of the team project.
+    :type project: str
+    :param detect: Automatically detect organization and project. Default is "on".
+    :type detect: str
+    """
+    organization, project = resolve_instance_and_project(
+        detect=detect, organization=organization, project=project)
+    client = get_build_client(organization)
+    folders = client.get_folders(path=path, project=project)
+    if len(folders) > 1:
```
gauravsaralMs: just curious, when this can happen?

azure-devops/azext_devops/dev/pipelines/_format.py
atbagga: I would still like to keep the start since the start...end will not be reliable to get anything useful. Lets close this during bug bash if people see any concern.

azure-devops/azext_devops/dev/pipelines/pipeline_folders.py
atbagga: Not required.

azure-devops/azext_devops/dev/pipelines/pipeline_folders.py
atbagga: This is because the List API documentation says -
image
but it is not working in that way so right now I am only getting one path which is the exact match.

Just to handle it better now I am checking against the list to find a matching path and update the matching path. Only error out if the matching path is not found.

atbagga:
Closing PR due to rebase issue.

Issue #774, opened 2019-09-03
[Question] Missing PR commands


thorhj:
According to [this page](https://docs.microsoft.com/en-us/cli/vsts/overview?view=vsts-cli-latest), the VSTS CLI is deprecated and replaced by this az extension instead. In [this file](https://github.com/Azure/azure-devops-cli-extension/blob/master/doc/command_mapping.md) the old VSTS commands are mapped to new commands in az.

Some of the commands from VSTS mentioned in this document seem to be missing. For instance, `vsts code pr complete` should be available as `az repos pr complete`, but it is not there. When I run `az repos pr -h` only the following commands are listed:

```
checkout
create
list
set-vote
show
update
```

Shouldn't all commands in the mappings document be available? Am I missing something?

azure-cli: **2.0.72**
azure-devops extension: **0.12.0**

atbagga:
@thorhj The **az repos pr update** command handles the missing commands like-
```
vsts code pr complete 
vsts code pr abandon
vsts code pr reactivate
```

though its **--status** parameter. See command documentation [here](https://docs.microsoft.com/en-us/cli/azure/ext/azure-devops/repos/pr?view=azure-cli-latest#ext-azure-devops-az-repos-pr-update).

Since these are just updates on the PR we consolidated these commands in azure-devops extension. 
Hopefully that answers your question. I will update the documentation accordingly.

Thanks for reporting the documentation mismatch.

thorhj:
Thanks @atbagga, just what I was looking for.

PR #784, opened 2019-09-10
Fix az repos pr command mapping

atbagga:

 - [ ] : This PR has a corresponding issue open in the Repository. Fixes: #774 


atbagga:
> should we add a banner in this md file telling that this mapping is valid for version 'X' so that we don't have to maintain this md file going forward?

@geverghe Please review this doc change.

PR #818, opened 2019-10-09
Help text changes to fix documentation

atbagga:

Fixes: https://github.com/Azure/azure-cli-extensions/issues/976

### DIFF
--- a/azure-devops/azext_devops/dev/artifacts/_help.py
+++ b/azure-devops/azext_devops/dev/artifacts/_help.py
@@ -10,7 +10,7 @@ def load_artifacts_help():
     helps['artifacts'] = """
         type: group
         short-summary: Manage Azure Artifacts.
-        long-summary:
+        long-summary: This command group is a part of the azure-devops extension.
     """
 
     helps['artifacts universal'] = """
--- a/azure-devops/azext_devops/dev/boards/_help.py
+++ b/azure-devops/azext_devops/dev/boards/_help.py
@@ -10,7 +10,7 @@ def load_boards_help():
     helps['boards'] = """
     type: group
     short-summary: Manage Azure Boards.
-    long-summary:
+    long-summary: This command group is a part of the azure-devops extension.
     """
 
     helps['boards work-item'] = """
--- a/azure-devops/azext_devops/dev/pipelines/_help.py
+++ b/azure-devops/azext_devops/dev/pipelines/_help.py
@@ -10,7 +10,7 @@ def load_pipelines_help():
     helps['pipelines'] = """
     type: group
     short-summary: Manage Azure Pipelines.
-    long-summary:
+    long-summary: This command group is a part of the azure-devops extension.
     """
 
     helps['pipelines runs'] = """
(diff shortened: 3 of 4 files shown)


atbagga:
> Doesn't look like the correct thing to do (having this information in long summary for commands).
> This problem should be common to all az extensions and should be solved at the meta level.
> 
> I am fine with workarounds as well but this is not looking correct

@gauravsaralMs  The fix is to share the knowledge that az pipelines command group is in azure-devops extension. This is hard to get today from the documentation page here - https://docs.microsoft.com/en-us/cli/azure/pipelines?view=azure-cli-latest. Please see linked issue for details. 
For me this is on same lines as to why we add az pipelines, az boards, az repos, and az artifacts help in the az devops help text. 

image


**Why other extensions don't need this -** Because they don't map to multiple top level command groups. They have 1:1 mapping like alias extension will introduce az alias.

I am open to suggestions on other ways to fix this.

atbagga:
Putting this on hold since there could be a better fix if Azure-CLI handles this in documentation auto-generation and there is an email conversation going on that topic with the concerned team.

geverghe:
Why are we putting this on hold - are we waiting for clarity on timelines before we push this out? Agree on the long term solution of doing it the right way till Azure CLI comes out with a meta solution to fix it for all extensions on their reference page/command experience in CLI, but till then this looks to be a good low cost fix to alleviate the pain. @gauravsaralMs - Do you anticipate any problems on releasing this and reverting to the Azure CLI mechanism post release?

gauravsaralMs:
@geverghe 
1. This creates an inconsistency in experience for the user, these 4 command groups can be are calling out which extension they are part of but rest all are not. So as a user after reading this I would assume batch command group is part of main azure cli and not part of any extension because it didn't called out.

2. reverting has a problem that user will first learn this method of discovering extension from documentation (which it self will be inconsistent) and then when a proper solution comes then it will a forget and learn for user.

3. Long summary is not meant for these kind of information. It should describe the functionality of command group and not give installation instructions because this same thing will show up in CLI (apart from documentation). 

Still feel free to go ahead with this if this looks good as a workaround.

geverghe:
Agree with 1 and 2. However, for 3 it is not a installation instruction but it is about the extension it belongs to because fundamentally we have a problem where a bunch of top level command groups are being bundled within the same extension where as others are pretty much 1:1 in terms of extension vs number of top level command groups exposed. Therefore, even in the documentation pointed out (command reference) people can fundamentally map back with the 1:1 implicit expectation set up.

However, the other element in the decision making process is the timeline for azure cli to come up with the overarching solution. Considering we are potentially the only extension with 4 top level groups within a single extension it is a bit more pronounced for us and therefore a bit more time crucial to prevent people running into the same error. If we are hoping for timeline confirmation and re-evaluation from mail then it make sense to hold till we get that. However, if we are not hoping for that from the mail/timeline is potentially Q2/Q3 next year, then it makes sense to put this up as a temporary fix understanding the risks that have been highlighted in 1 and 2.

PR #835, opened 2019-10-15
Re-adding optional project parameter to artifacts universal commands

salvmd:
The required change in ArtifactTool has now been released.

This PR has the same changes as #785 , which was reverted in #789 since the change in ArtifactTool wasn't ready at the time.

### DIFF
--- a/azure-devops/azext_devops/dev/artifacts/universal.py
+++ b/azure-devops/azext_devops/dev/artifacts/universal.py
@@ -14,7 +14,15 @@ from azext_devops.dev.common.external_tool import ProgressReportingExternalToolI
 logger = get_logger(__name__)
 
 
-def publish_package(feed, name, version, path, description=None, scope='organization', organization=None, project=None, detect=None):
+def publish_package(feed,
+                    name,
+                    version,
+                    path,
+                    description=None,
+                    scope='organization',
+                    organization=None,
+                    project=None,
+                    detect=None):
     """Publish a package to a feed.
     :param scope: Scope of the feed: 'project' if the feed was created in a project, and 'organization' otherwise.
     :type scope: str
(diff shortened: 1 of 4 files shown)


azure-devops/azext_devops/dev/artifacts/universal.py
```diff
@@ -27,12 +27,18 @@ def publish_package(feed, name, version, path, description=None, organization=No
     :type path: str
     """
     colorama.init()   # Needed for humanfriendly spinner to display correctly
-    organization = resolve_instance(detect=detect, organization=organization)
+
+    organization, project = resolve_instance_and_project(
+        detect=detect,
+        organization=organization,
+        project=project,
```
atbagga: This will automatically pick up the project from default config or local git repo context and target call will always be made at project level even if he user did not intend to make it that way.

What is the difference in targeting a feed with or without the project name? 

Is there a use case that a user will want to not input the project name? If yes, this command will now not allow the user to do it since it will pick up the project from auto detect logic.

pspill:
> This will automatically pick up the project from default config or local git repo context and target call will always be made at project level even if he user did not intend to make it that way.
> 
> What is the difference in targeting a feed with or without the project name?
> 
> Is there a use case that a user will want to not input the project name? If yes, this command will now not allow the user to do it since it will pick up the project from auto detect logic.

@atbagga This issue is fixed, by using in addition the optional scope parameter:
- The default scope is organization, and prior to this PR project scoped was not allowed.  Therefore, existing scripts targeting org scoped feeds won't need to change.
- if "--project <p>" is supplied, then the user must also supply "--scope project"
- if a default project is configured, then it will only be used if the user supplies the "--scope project" parameter.

pspill:
> Please add tests. They are bare minimum unit tests but still help sometimes.
> 
> https://github.com/Azure/azure-devops-cli-extension/blob/26f69e71a18cb5c7984d565ffbd32acda6eb4000/azure-devops/azext_devops/test/artifacts/test_univeral.py#L49

@atbagga I don't have permissions to modify @salvmd 's fork, so I create a new PR that includes this change + unit tests.  We can either take this change and then mine, or just take mine, either way.  PR #862

gauravsaralMs:
> > Please add tests. They are bare minimum unit tests but still help sometimes.
> > https://github.com/Azure/azure-devops-cli-extension/blob/26f69e71a18cb5c7984d565ffbd32acda6eb4000/azure-devops/azext_devops/test/artifacts/test_univeral.py#L49
> 
> @atbagga I don't have permissions to modify @salvmd 's fork, so I create a new PR that includes this change + unit tests. We should abandon this PR and take #862 instead.

Closing this PR as per this comment

PR #838, opened 2019-10-17
Create CODEOWNERS

atbagga:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,4 @@
-# This is a comment.
-# Each line is a file pattern followed by one or more owners.
-# https://help.github.com/en/articles/about-code-owners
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,11 @@
+# This is a comment.
```
geverghe: This is more a query than any recommendation
Should we keep all this default commentary from the code owner file? Not that it is hurting anything; but still.  

```
# https://help.github.com/en/articles/about-code-owners

# These owners will be the default owners for everything in
# the repo. Unless a later match takes precedence,
*       @atbagga @gauravsaralMs @ishitam8

# These owners own any file in the `/docs`
# directory in the root of your repository.
/doc/  @geverghe*/```

PR #840, opened 2019-10-21
Fix merge pipeline failure and improve debugging of backCompat script

atbagga:

### DIFF
--- a/scripts/backCompatChecker.py
+++ b/scripts/backCompatChecker.py
@@ -55,8 +55,16 @@ def extractArgumentsFromCommand(command):
 
     return argumentList
 
+#Check the installed extensions
+subprocess.run(['az', 'extension', 'list'], shell=True, stdout=subprocess.PIPE)
+
+# remove azure-devops extension from index (if installed)
+subprocess.run(['az', 'extension', 'remove', '-n', 'azure-devops'], shell=True, stdout=subprocess.PIPE)
+
 # install extension from index
 subprocess.run(['az', 'extension', 'add', '-n', 'azure-devops'], shell=True, stdout=subprocess.PIPE)
+
+subprocess.run(['az', 'extension', 'list'], shell=True, stdout=subprocess.PIPE)
 
 # add extension path to sys.path so that we can get all the commands
 import sys
@@ -86,6 +94,8 @@ def findExtension():
         for file in f:
             if file.endswith('.whl'):
                 return os.path.join(p, file)
+
+print('Install extension (loaded from current code).')
 
 newExtensionLocation = findExtension()
 subprocess.run(['az', 'extension', 'add', '--source', newExtensionLocation, '-y'], shell=True, stdout=subprocess.PIPE)
--- a/scripts/backCompatChecker.py
+++ b/scripts/backCompatChecker.py
@@ -55,17 +55,11 @@ def extractArgumentsFromCommand(command):
 
     return argumentList
 
-# Check the installed extensions
-subprocess.run(['az', 'extension', 'list'], shell=True)
-
 # remove azure-devops extension from index (if installed)
 subprocess.run(['az', 'extension', 'remove', '-n', 'azure-devops'], shell=True)
 
 # install extension from index
 subprocess.run(['az', 'extension', 'add', '-n', 'azure-devops'], shell=True)
-
-subprocess.run(['az', 'devops', 'security', 'permission', '-h'], shell=True)
-subprocess.run(['az', 'devops', 'security', 'permission', 'reset', '-h'], shell=True)
 
 # Check the installed extensions
 subprocess.run(['az', 'extension', 'list'], shell=True)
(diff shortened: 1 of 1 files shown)


scripts/backCompatChecker.py
```diff
@@ -85,21 +93,27 @@ def findExtension():
...
 subprocess.run(['az', 'extension', 'add', '--source', newExtensionLocation, '-y'], shell=True, stdout=subprocess.PIPE)
 
+# Check the installed extensions
+subprocess.run(['az', 'extension', 'list'], shell=True)
+
 # get a set of old commands, we are not reusing the set from ext because we want to keep this clean
 oldCommands = []
 for oldArgument in oldArguments:
     if oldArgument.command not in ignoreCommands:
-        if not any(oldArgument.command in s for s in oldCommands):
```
atbagga: The check fails for command = 'devops security permission reset' list containing 'devops security permission reset-all' and thus all params of command are being detected as missing. 

Still not sure how it started failing without any code change.. did the python behavior change?

Also since the preview mechanism is changed these commands are the ones which were previously not even getting checked for back compat but are now because the 'PREVIEW' check in this script will not work.

atbagga: Rest of the changes in the file are only to improve the debugging from logs.

ishitam8: I agree, how did it fail all of a sudden since, the python version(3.7) and azure-cli version(2.0.75) looks same on recent and earlier pipelines. 
But somehow, the command PREVIEW tags are still present in backCompactChecker script logs, even though, azure-cli version 2.0.75 already has the extension PREVIEW feature inbuilt.

atbagga: > I agree, how did it fail all of a sudden since, the python version(3.7) and azure-cli version(2.0.75) looks same on recent and earlier pipelines.
> But somehow, the command PREVIEW tags are still present in backCompactChecker script logs, even though, azure-cli version 2.0.75 already has the extension PREVIEW feature inbuilt.

That is because with new preview mechanism the text contains 'preview' and not 'PREVIEW' which is what we check in the script to exclude those commands.

atbagga: Updating here for completing the thread... As discussed offline.. 

The issue here was were running this test against the already installed version of extension in the agent which was 0.12.0 and preview commands were not getting tested because they are excluded by the script. 
**FIX 1: This is fixed by adding a extension remove command at start of script.**

The agent got updated and the extension version 0.13.0 was available where the PREVIEW check failed to block the commands and this started failing due to a code issue which was always there which returns positive match for list membership even on substring match. 
**FIX 2: Check for preview according to new code.**
**FIX 3: Substring match issue in list membership check.**

scripts/backCompatChecker.py
```diff
@@ -55,8 +55,14 @@ def extractArgumentsFromCommand(command):
 
     return argumentList
 
+# remove azure-devops extension from index (if installed)
+subprocess.run(['az', 'extension', 'remove', '-n', 'azure-devops'], shell=True)
```
atbagga: Fix 1: This will make sure we test against the latest extension version in the index.

scripts/backCompatChecker.py
```diff
@@ -35,7 +35,7 @@ def extractArgumentsFromCommand(command):
     help_text = subprocess.run(commandExtended.split(' '), shell=True, stdout=subprocess.PIPE)
     print('help text for ' + command)
     print(help_text)
-    if "PREVIEW" in str(help_text):
+    if " is in preview" in str(help_text):
```
atbagga: FIX 2: Exclude preview commands as per new preview mechanism.

scripts/backCompatChecker.py
```diff
@@ -85,21 +93,27 @@ def findExtension():
...
 
+# Check the installed extensions
+subprocess.run(['az', 'extension', 'list'], shell=True)
+
 # get a set of old commands, we are not reusing the set from ext because we want to keep this clean
 oldCommands = []
 for oldArgument in oldArguments:
     if oldArgument.command not in ignoreCommands:
-        if not any(oldArgument.command in s for s in oldCommands):
+        if not (oldArgument.command in oldCommands):
```
atbagga: FIX 3:  Substring match issue in list membership check. The check fails for command = 'devops security permission reset' list containing 'devops security permission reset-all' and thus all params of command are being detected as missing.

scripts/backCompatChecker.py
gauravsaralMs: will this throw if extension is not present?

atbagga: Its is subprocess so script will not fail. I Just tested it here - https://dev.azure.com/ms/azure-devops-cli-extension/_build/results?buildId=44027&view=logs&j=543512b3-d0ca-5e9c-ec0b-164b21561d9e&t=4cdb99d1-90fe-5988-0df0-0605d016b7e4&l=23

Issue #801, opened 2019-09-25
[Feature Request] Document the use of AZURE_DEVOPS_EXT_GITHUB_PAT


ams0:
the `az devops service-endpoint github create` command can be run non-interactively by setting a `AZURE_DEVOPS_EXT_GITHUB_PAT` environment variable, but it's not documented (only in code [here](https://github.com/Azure/azure-devops-cli-extension/blob/ae2d49831f11601e44b30257e9d23334638fd770/azure-devops/azext_devops/dev/common/const.py#L19)).

Please add a documentation to the command.

geverghe:
@gauravsaralMs - isn't this done (code complete but yet to release)? Including link to [this](https://docs.microsoft.com/en-us/azure/devops/cli/service_endpoint?view=azure-devops) document in service connection create command help?

geverghe:
@gauravsaralMs - further, if it is done, could  you expand scope to include the proposal mentioned in #698

Issue #698, opened 2019-06-28
Include capability to provide the service endpoint service principal key as an environment variable


geverghe:
It would be great if the docs included this possibility for providing the service endpoint service principal key as an environment variable, in order for it to be able to run in an automated way without a need for user prompt.

geverghe:
Include the following for each of az devops service-endpoint create commands:

1. az devops service-endpoint create -h
Command:   az devops service-endpoint create : Create a service endpoint using configuration file. You can learn more about this [here](https://docs.microsoft.com/en-us/azure/devops/cli/service_endpoint?view=azure-devops#create-service-endpoint-using-configuration-file). 

*please note the link with the reference on the appropriate title

2. az devops service-endpoint azurerm create : Create an Azure RM type service endpoint. For automation, please set service principal password/secret in AZURE_DEVOPS_EXT_AZURE_RM_SERVICE_PRINCIPAL_KEY environment variable. You can learn more about this [here](https://docs.microsoft.com/en-us/azure/devops/cli/service_endpoint?view=azure-devops#creating-azure-rm-service-endpoint). 

3. az devops service-endpoint github create: Create a GitHub service endpoint. For automation purpose set the GitHub PAT token in AZURE_DEVOPS_EXT_GITHUB_PAT environment variable. You can learn more about this [here](https://docs.microsoft.com/en-us/azure/devops/cli/service_endpoint?view=azure-devops#creating-github-service-endpoint).

PR #843, opened 2019-10-22
Documentation fix for service endpoint commands

atbagga:

 - [ ] : This PR has a corresponding issue open in the Repository. #801 #698 

### DIFF
--- a/azure-devops/azext_devops/dev/team/_help.py
+++ b/azure-devops/azext_devops/dev/team/_help.py
@@ -28,14 +28,35 @@ def load_team_help():
     short-summary: Manage service endpoints/connections
     """
 
+    helps['devops service-endpoint create'] = """
+    type: command
+    short-summary: Create a service endpoint using configuration file.
+    long-summary: You can learn more about this at https://aka.ms/azure-devops-cli-service-endpoint
+    """
+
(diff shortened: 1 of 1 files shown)


azure-devops/azext_devops/dev/team/_help.py
```diff
@@ -28,16 +28,37 @@ def load_team_help():
...
+
     helps['devops service-endpoint github'] = """
     type: group
     short-summary: Manage GitHub service endpoints/connections
     """
 
+    helps['devops service-endpoint github create'] = """
+    type: command
+    short-summary: Create a GitHub service endpoint.
+    long-summary: For automation purpose set the GitHub PAT token in AZURE_DEVOPS_EXT_GITHUB_PAT
```
gauravsaralMs: v.v. nit: maybe the wordings in both should match.

atbagga: @geverghe  To consider the feedback.
+ If we need to update section reference in the link.. in that case we will need 3 different aka.ms links.

geverghe: 1.Go ahead with point 1.
2. Didn't quite get point 2; why do we need 3 aka.ms links?

PR #846, opened 2019-10-23
Deprecate feedback command

atbagga:

- Hide command
- Deprecate message pointing to use az feedback instead.
- Do not open the url only print the message as it used to do earlier (so that the hit count on the url comes down.)

image

### DIFF
--- a/azure-devops/azext_devops/dev/team/feedback.py
+++ b/azure-devops/azext_devops/dev/team/feedback.py
@@ -5,7 +5,6 @@
 # --------------------------------------------------------------------------------------------
 
 from __future__ import print_function
-from azext_devops.dev.common.utils import open_url
 
 
 def feedback():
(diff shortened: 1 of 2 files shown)


azure-devops/azext_devops/dev/team/feedback.py
```diff
@@ -14,4 +13,3 @@ def feedback():
     url = 'https://aka.ms/azure-devops-cli-feedback'
     print('Thank you for taking the time to share your feedback. Please submit your feedback on the following web ' +
           'page: {url}'.format(url=url))
-    open_url(url=url)
```
gauravsaralMs: not needed actually :) but no harm in removing as well

atbagga: Reason for removing: Since it auto opens the number of hits to the URL will be high. By doing this I am hoping the hits will come down significantly and then we can safely remove the command.

PR #862, opened 2019-10-31
Re-adding optional project parameter to artifacts universal commands + unit tests

pspill:
The required change in ArtifactTool has now been released.

This adds unit tests missing in #835, so contains the changes in #835 + unit tests.  Otherwise, this PR has the same changes as #785 , which was reverted in #789 since the change in ArtifactTool wasn't ready at the time.


gauravsaralMs:
close open to re-trigger pipelines

PR #863, opened 2019-11-01
use python 3.7 instead of latest because 3.8 has issues

gauravsaralMs:


gauravsaralMs:
force merging this for now, so that pipelines go back to green (we can revert this when cli edge get the fix in https://github.com/Azure/azure-cli/issues/11041

PR #880, opened 2019-11-13
Fallback to file storage when keyring is not working

### LINKED ISSUE #762: az devops login fails with "InitError: Failed to create the collection: Prompt dismissed.."
## Describe the bug

**Command Name**
`az devops login
Extension Name: azure-devops. Version: 0.12.0.`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:


Failed to create the collection: Prompt dismissed..
Traceback (most recent call last):
  File "/usr/lib64/az/lib/python2.7/site-packages/knack/cli.py", line 206, in invoke
    cmd_result = self.invocation.execute(args)
  File "/usr/lib64/az/lib/python2.7/site-packages/azure/cli/core/commands/__init__.py", line 603, in execute
    raise ex
InitError: Failed to create the collection: Prompt dismissed..
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- `az devops login`

## Expected Behavior

Perform a successful login

## Environment Summary
```
Linux-5.0.16-100.fc28.x86_64-x86_64-with-fedora-28-Twenty_Eight
Python 2.7.15
Shell: bash

azure-cli 2.0.71

Extensions:
azure-devops 0.12.0

```

ishitam8:

Fixes #762 

If keyring is not configured properly, we will store the credentials in file (in user home directory) with a warning

KeyringFailedSnap

### DIFF
--- a/azure-devops/azext_devops/dev/common/credential_store.py
+++ b/azure-devops/azext_devops/dev/common/credential_store.py
@@ -55,6 +55,7 @@ class CredentialStore:
             import keyring
         except ImportError:
             return None
+        token = None
         try:
             token = keyring.get_password(key, self._USERNAME)
         except Exception as ex:  # pylint: disable=broad-except
--- a/azure-devops/azext_devops/dev/common/credential_store.py
+++ b/azure-devops/azext_devops/dev/common/credential_store.py
@@ -37,9 +37,9 @@ class CredentialStore:
         except Exception as ex:  # pylint: disable=broad-except
             # store credentials in azuredevops config directory if keyring is missing or malfunctioning
             if sys.platform.startswith(self._LINUX_PLATFORM):
-                logger.warning('Failed to store PAT using keyring; falling back to file storage.\
-                    You can clear the stored credential by running az devops logout.\
-                    Refer https://aka.ms/azure-devops-cli-auth to know more on sign in with PAT.')
+                logger.warning('Failed to store PAT using keyring; falling back to file storage.')
+                logger.warning('You can clear the stored credential by running az devops logout.')
+                logger.warning('Refer https://aka.ms/azure-devops-cli-auth to know more on sign in with PAT.')
                 logger.debug('Keyring failed. ERROR :%s', ex)
                 logger.debug('Storing credentials in the file: %s', self._PAT_FILE)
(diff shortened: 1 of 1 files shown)


azure-devops/azext_devops/dev/common/credential_store.py
```diff
@@ -79,9 +82,11 @@ def clear_password(self, key):
             keyring.delete_password(key, self._USERNAME)
```
gauravsaralMs: may be a corner case, but what if set has issues and delete does not have issue.
In that case customer will always see `The credential was not found`.

Should we instead look in both places and clean?

azure-devops/azext_devops/dev/common/credential_store.py
```diff
@@ -54,11 +56,12 @@ def get_password(self, key):
 
         try:
             return keyring.get_password(key, self._USERNAME)
```
gauravsaralMs: may be a corner case, but what if set has issues and get does not have issue.
With this change should we start looking in both the places for password?

azure-devops/azext_devops/dev/common/credential_store.py
ishitam8: Will check in both keyring and file for
1. get password
2. delete password

azure-devops/azext_devops/dev/common/credential_store.py
```diff
@@ -33,10 +33,14 @@ def set_password(self, key, token):
                 keyring.delete_password(key, self._USERNAME)
```
atbagga: We are not clearing password from file here. But I cannot think of a scenario where this would have any impact so looks ok. I am still mentioning this to think through any scenario which could cause concern.

It could just end up having credentials stored in both file and keyring (steps below) but it should be taken care by logout.

Steps:
1. Keyring dependency is missing and user runs az devops login
2. Keyring issue gets sorted out.
3. User runs az  devops login without running logout.

ishitam8: I agree, but i think that's fine since we are giving explicit warning that we are storing credentials in file and asking them to logout later.

PR #889, opened 2019-11-18
Adding step in release new version to fix released version pipeline

atbagga:


doc/release_new_version.md
```diff
@@ -38,3 +38,7 @@ Also update build pipelines YAMLS
 Update [AzureDevOpsCli-Released-Version-Check](https://dev.azure.com/ms/azure-devops-cli-extension/_build?definitionId=36) to run from the branch which was used to do the release.
```
ishitam8: May be remove these lines?
Since **released version check** was renamed to **release-version** ,
and both text says the same thing.

Issue #817, opened 2019-10-09
[Feature Request] service-endpoint create needs option to specify pipeline policies


jfpanisset:
When you create a service connection via the GUI, you get a check box to specific "Allow all pipelines to use this connection". This causes not only a POST connection to be sent to the `_apis/serviceendpoint/endpoints` REST API, but also a PATCH request to `_apis/build/authorizedresources` containing:

```
[{"authorized":true,"id":"the-resource-id","name":"the-resource-name","type":"endpoint"}]
```

If instead you use the CLI to create a service connection using:

```
az devops service-endpoint create ----service-endpoint-configuration foo.json
```

the PATCH request is not sent, and by default the service connection cannot be accessed by the pipeline until you enable that manually in the GUI. There may be a way to manipulate this permission using the az CLI (and if there is I'd love to know about it), it would be very convenience to add a command line option named something like `--all-pipelines-use` which would default the access to the service connection to all pipelines in the project. Even better would be explicit CLI support to manipulate access from specific pipelines if required.

atbagga:
The policy cannot be at pipeline level since that is not supported. It is either all pipelines or none.

Till the command is enhanced to support the option you can use this command to change that setting from the CLI (complete your automation flow).

```az devops invoke --http-method patch --area build --resource authorizedresources --debug --route-parameters project=atbagga --api-version 5.0-preview --in-file .\a.txt --encoding ascii```

Where a.txt will contain the json that you mentioned in the issue. 

```[{"authorized":true,"id":"the-resource-id","name":"the-resource-name","type":"endpoint"}]```

jfpanisset:
Thank you very much, `az devops invoke` seems to work and correctly sets the "Allow all pipelines to use this service connection" property for my newly created Docker Hub service connection. Before I got your answer I was trying to use `az rest --method patch ...` but wasn't getting anywhere.

Until 'az devops service-endpoint create` can be enhanced with this functionality, would it make sense to add this recommendation to the documentation?

geverghe:
Can we investigate what it means to enable this in the command? @jfpanisset - Feel free to chime in based on your expectation. 
### 1.  `az devops service-endpoint create --enable-for-all`
Input - service end point name and other required inputs based on service endpoint type
Output - json output of the create service endpoint connection, 
Advantages 
1. One command to achieve objective

Drawback
1. Cannot understand what happened to the status of enable for all; did it successfully happen? was there an error? 
[...]

PR #897, opened 2019-11-26
Service Endpoint update command to set pipeline access property

atbagga:

 - [ ] : This PR has a corresponding issue open in the Repository. Fixes: #817 

### DIFF
--- a/azure-devops/azext_devops/dev/team/service_endpoint.py
+++ b/azure-devops/azext_devops/dev/team/service_endpoint.py
@@ -8,6 +8,7 @@ from __future__ import print_function
 import os
 from knack.log import get_logger
 from knack.prompting import prompt_pass
+from knack.util import CLIError
 from azext_devops.devops_sdk.v5_0.service_endpoint.models import ServiceEndpoint, EndpointAuthorization
 from azext_devops.dev.common.services import get_service_endpoint_client, resolve_instance_and_project
 from azext_devops.dev.common.const import CLI_ENV_VARIABLE_PREFIX, AZ_DEVOPS_GITHUB_PAT_ENVKEY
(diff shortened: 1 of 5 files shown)


azure-devops/azext_devops/dev/team/service_endpoint.py
```diff
@@ -172,3 +199,27 @@ def create_service_endpoint(service_endpoint_configuration,
...
+    se = client.get_service_endpoint_details(project, id)
+
+    # set authorization if get service endpoint succeeded
+    from azext_devops.dev.pipelines.pipeline_utils import set_authorize_resource, get_authorize_resource
+    set_authorize_resource(
+        authorized=enable_for_all, res_id=se.id, name=se.name, res_type='endpoint',
+        organization=organization, project=project)
+
+    authorized = get_authorize_resource(res_id=se.id, res_type='endpoint', organization=organization, project=project)
+    return ServiceEndpointAuthorized(service_endpoint_parameters=se, authorized=authorized)
```
ishitam8: Why are we doing this?
Can't we just return response from authorize_project_resources ?
Since, this is a separate command for enable-all

Also, if its really necessary can we move ServiceEndpointAuthorized to some helper class?

azure-devops/azext_devops/dev/team/service_endpoint.py
```diff
@@ -172,3 +199,27 @@ def create_service_endpoint(service_endpoint_configuration,
...
+
+
+def update_service_endpoint(id, enable_for_all=None, organization=None,  # pylint: disable=redefined-builtin
+                            project=None, detect=None):
+    """Update a service endpoint
+    :param id: ID of the service endpoint.
+    :type id: str
+    """
+    if enable_for_all is None:
+        raise CLIError('Atleast one property to be updated must be specified.')
```
ishitam8: 'enable_for_all' is mandatory ? [Since its the only argument]

atbagga: Future proof... we will end up adding new params to the command and all should remain optional. 

There is error handling right now if it is not set.

azure-devops/azext_devops/dev/team/service_endpoint.py
atbagga: Why a helper class? this is not used anywhere else except with service endpoint commands.

The command is service-endpoint update so it should return details of service endpoint also.. again this is to allow adding more params in future to this command to update actual service endpoint metadata.

atbagga: As discussed offline.. The only concern remains on maintenance overhead of JSON class being referenced here which should be caught in linting check if it changes during client upgrade. 

Let me know if there are any more concerns.

Issue #776, opened 2019-09-03
az artifacts universal download fails with Access denied and file not found


wpschaub:
## Describe the bug

**Command Name**
`az artifacts universal download
Extension Name: azure-devops. Version: 0.12.0.`

**Errors:**

First error when using verbose logging:
```
An error occurred while extracting the Universal Packages tooling: [WinError 5] Access is denied: 'C:\\Users\\WS19312\\.azure\\azuredevops\\cli\\tools\\artifacttool\\a445ec82-2166-4964-946e-d868ead0c4e3' -> 'C:\\Users\\WS19312\\.azure\\azuredevops\\cli\\tools\\artifacttool\\ArtifactTool_win10-x64_0.2.104'
```

Error trace reported by tool:
```
[WinError 2] The system cannot find the file specified
Traceback (most recent call last):
pip-install-76yvurxi\knack\knack\cli.py, ln 206, in invoke
cli\core\commands\__init__.py, ln 575, in execute
cli\core\commands\__init__.py, ln 633, in _run_jobs_serially
cli\core\commands\__init__.py, ln 624, in _run_job
azext_devops\dev\common\exception_handler.py, ln 26, in azure_devops_exception_handler
    reraise(*sys.exc_info())
Temp\pip-install-76yvurxi\six\six.py, ln 693, in reraise
cli\core\commands\__init__.py, ln 603, in _run_job
...
Azure\CLI2\lib\subprocess.py, ln 709, in __init__
    restore_signals, start_new_session)
Azure\CLI2\lib\subprocess.py, ln 997, in _execute_child
    startupinfo)
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

## To Reproduce:
Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

- `az login`
- `az extension add --name azure-DevOps`
- `az artifacts universal download --organization {} --feed {} --name {} --version {} --path {} --debug`

## Environment Summary
```
Windows-10-10.0.16299-SP0
Python 3.6.6
Shell: powershell.exe

azure-cli 2.0.68 *

Extensions:
azure-devops 0.12.0

```

shubham90:
@wpschaub Do you have write permission to "C:\\Users\\WS19312"? Can you try running the command in an administrators prompt to eliminate the possibility of permissions issue?

wpschaub:
I have write permissions to the folder and have tried running it in an administrators prompt.

[...]

PR #925, opened 2019-12-13
Retrying os rename on failure and more logging

shubham90:

 - [ Issue #776 ] : This PR has a corresponding issue open in the Repository.

This PR is adding more logging and retry around os.rename which is failing for some customers because some other process is locking the file(may be an antivirus). There shouldn't be any retry in case of success.

### DIFF
--- a/azure-devops/azext_devops/dev/common/artifacttool_updater.py
+++ b/azure-devops/azext_devops/dev/common/artifacttool_updater.py
@@ -9,6 +9,7 @@ import platform
 import shutil
 import stat
 import sys
+import time
 import uuid
 import zipfile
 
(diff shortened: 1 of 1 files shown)


azure-devops/azext_devops/dev/common/artifacttool_updater.py
```diff
@@ -124,9 +125,22 @@ def _update_artifacttool(uri, release_id):
...
+                # number of times to retry
+                retries = 10
+                for _ in range(retries - 1):
+                    try:
+                        os.rename(release_temp_dir, release_dir)
+                        break
+                    except BaseException as ex:
+                        logger.debug("An error occurred while renaming the Universal Packages tooling: %s. Retrying...", ex)
+                        time.sleep(1)
+                else:
```
atbagga: Hanging else?

azure-devops/azext_devops/dev/common/artifacttool_updater.py
```diff
@@ -124,9 +125,22 @@ def _update_artifacttool(uri, release_id):
 
             # Move the release into the real releases location
             release_dir = _compute_release_dir(release_id)
-            logger.debug("Moving downloaded ArtifactTool from %s to %s", release_temp_dir, release_dir)
-            os.rename(release_temp_dir, release_dir)
-            logger.info("Downloaded Universal Packages tooling successfully")
+            if os.path.exists(release_dir):
```
atbagga: Can it end up in a situation where directory exists but it is empty. The command will not even try to put the tool in the directory in that case.

azure-devops/azext_devops/dev/common/artifacttool_updater.py
shubham90: This 'else' is for the FOR loop to retry one last time before failing. Doing it this way avoids adding the same logic in except block.

azure-devops/azext_devops/dev/common/artifacttool_updater.py
shubham90: That shouldn't happen unless the download of the tool itself fails, in which case this will fail much earlier.

azure-devops/azext_devops/dev/common/artifacttool_updater.py
```diff
@@ -124,9 +125,25 @@ def _update_artifacttool(uri, release_id):
...
+                logger.info(
```
salvmd: Talked offline: let's make this a warning since it's an unexpected case.

Issue #950, opened 2020-02-14
[Bug] Unable to disable telemetry


aodj:
**Describe the bug**
It's not possible to disable the telemetry collection within the azure-devops extension for two separate reasons:
* When setting the configuration using `az configure -d collect_telemetry=...` the value is saved under the `[defaults]` heading in the azure-cli config file, which isn't detected by azure-devops since it's only reading the `[core]` section: [relevant link](https://github.com/Azure/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/dev/common/telemetry.py#L68-L69)
* Secondly, the [documentation](https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration?view=azure-cli-latest#cli-configuration-file) around the `az configure` command states that
 > Booleans are case-insensitive, and are represented by the following values.
 > * True: 1, yes, true, on
 > * False: 0, no, false, off

which means that setting `collect_telemetry=false` is the same as setting `collect_telemetry=no`. However, the code in azure-devops does an equivalence check against `no` rather than using the `getboolean()` method that comes with the Python ConfigParser module for these situations. [relevant link](https://github.com/Azure/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/dev/common/telemetry.py#L70)

**To Reproduce**
```bash
$ az --version
azure-cli                         2.0.81

command-modules-nspkg              2.0.3
core                              2.0.81
nspkg                              3.0.4
telemetry                          1.0.4

Extensions:
azure-devops                      0.17.0

Python location '/usr/bin/python3'
Extensions directory '/home/ansible/.azure/cliextensions'

Python (Linux) 3.6.8 (default, Aug  7 2019, 17:28:10)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]

Legal docs and information: aka.ms/AzureCliLegal



Your CLI is up-to-date.

Please let us know how we are doing: https://aka.ms/clihats
```

1. Disable the `collect_telemetry` flag in azure-cli with `az configure -d collect_telemetry=false`
2. Run an azure-devops specific command, with the `--verbose` and `--debug` flags set f.ex: `az pipelines build list -o table --verbose --debug`
3. Note the logged output: `Azure devops telemetry enabled.`
4. Reconfigure the azure-cli with `az configure -d collect_telemetry=no`
5. Rerun the `az pipelines ...` command, noting the logged output still contains `Azure devops telemetry enabled.`
[...]

PR #951, opened 2020-02-17
Telemetry Config Check Bug Fix

mitsha-microsoft:

Fix for the issue #950

### DIFF
--- a/azure-devops/azext_devops/dev/common/telemetry.py
+++ b/azure-devops/azext_devops/dev/common/telemetry.py
@@ -67,7 +67,9 @@ def _is_telemetry_enabled():
     az_config = az_cli_ctx.config
     if az_config.has_option('core', 'collect_telemetry'):
         collect_telemetry = az_config.get('core', 'collect_telemetry')
-    return bool(collect_telemetry is None or collect_telemetry != 'no')
+    elif az_config.has_option('defaults', 'collect_telemetry'):
+        collect_telemetry = az_config.get('defaults', 'collect_telemetry')
+    return bool(collect_telemetry is None or collect_telemetry not in ('no', 'false', 'off', '0'))
 
 
 def _try_send_tracking_ci_event_async(organization=None):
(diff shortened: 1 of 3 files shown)


azure-devops/azext_devops/dev/common/telemetry.py
```diff
@@ -67,7 +67,9 @@ def _is_telemetry_enabled():
     az_config = az_cli_ctx.config
     if az_config.has_option('core', 'collect_telemetry'):
         collect_telemetry = az_config.get('core', 'collect_telemetry')
-    return bool(collect_telemetry is None or collect_telemetry != 'no')
+    elif az_config.has_option('defaults', 'collect_telemetry'):
+        collect_telemetry = az_config.get('defaults', 'collect_telemetry')
```
atbagga: I think the [documentation](https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration?view=azure-cli-latest#cli-configuration-values-and-environment-variables) suggests default should overwrite core. 
This code would do the reverse. 

This change is still prone to mismatch later if Azure CLI changes their logic or priority. The telemetry check should be given by Azure CLI as a public function to be consumed by the extensions so that the logic remains in one place. 

Lets follow up on this with Azure CLI before pushing this fix.

azure-devops/azext_devops/dev/common/telemetry.py
```diff
@@ -67,7 +67,9 @@ def _is_telemetry_enabled():
...
+    return bool(collect_telemetry is None or collect_telemetry not in ('no', 'false', 'off', '0'))
```
aodj: Would it not make sense to use the `getboolean` method here? [Link](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.getboolean).

I'm the guy who originally reported that the telemetry disabling wasn't working, and I've no insight into this config object, so if it turns out to not be a `ConfigParser` object, and I'm way off base I apologize. The reason I bring it up is that I saw in one of the other Azure CLI repos, that `getboolean` was being called.

atbagga: @aodj Your feedback is valid in terms of making the check consistent. 
We are still working on getting this fixed in the correct way, we will mostly work with Azure CLI or contribute there to have the right fix in place.

mitsha-microsoft:
Abandoning this PR to create a new one

Issue #1046, opened 2020-09-07
[Feature Request] Create a pipeline with yaml from another repository


jarpsimoes:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I cannot create a project..
When we needs create a pipeline with AZ CLI with yaml present on another repository, as available on UI, now the CLI return error 404 if try with that command:

`pipelines create --name <pipeline_name> --branch master --org <organization url> --project <project_name> --repository <repository_pipelines_project>/<repository_name> --repository-type tfsgit --yaml-path /pipeline.yaml
`

**Use case:**
All pipelines in a project and each other projects will re-use that pipelines

**Describe the solution you'd like**
On method client.py:_create_request_message when exist 'repositoryId' on route_values and 'repositoryId' contains '/' build url to get sources from other project.

PR #1047, opened 2020-09-08
ADD: Feature 1046

jarpsimoes:
This PR has a corresponding issue: Create a pipeline with yaml from another repository #1046

### DIFF
--- a/azure-devops/azext_devops/devops_sdk/client.py
+++ b/azure-devops/azext_devops/devops_sdk/client.py
@@ -9,6 +9,7 @@ import logging
 import os
 import re
 import uuid
+import urllib.parse
 
 from msrest import Deserializer, Serializer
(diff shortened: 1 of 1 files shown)


ashishonce:
@narula0781  for visibility. 
this looks like changes the current basic nature of the command. Will have to decide was this by design or an actual bug, to restrict pipeline files from same repo. 
@romil07 thoughts?

romil07:
I don't have a direct answer to this. We should ask PM.
But I tried creating a classic release pipeline and we support generic git repos for that. I couldn't find a way to do the same for yml.

Steg17 (approved):

olljanat:
@ashishonce @romil07 two years later I wonder that was there some decission after all if this will be accepted or not? And if answer is not I would like to userstand that why? On UI that feature anyway exists.

PR #1102, opened 2021-02-12
feat(boards): Allow setting remote related and other WI URL relations directly

asmundg:
The `boards work-item add relation` command requires looking up related
work items by ID before adding their relation URL. This does not support
relationships such as 'Remote Related', where we cannot look up remote
work item id's.

To support the 'Remote Related' scenario (or scenarios involving non-work
item URLs, as in https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/work%20items/update?view=azure-devops-rest-6.0#add-a-hyperlink), we add a new command, `add-url`,
which bypasses the work item id lookup, and sets the relation URL directly.

 - [x] : This PR has a corresponding issue open in the Repository: #1109 


asmundg:
@ashishonce, @narula0781 any chance you can take a look at this?

Issue #1151, opened 2021-07-13
add a warning when calls go to Server.


gauravsaralMs:
Azure DevOps has 2 offering

- [Azure DevOps service](https://azure.microsoft.com/en-in/services/devops/)
- [Azure DevOps server](https://azure.microsoft.com/en-in/services/devops/server/)

This CLI offering is supported against Azure DevOps service but **not** against Azure DevOps server.
The Goal of this feature is to add a warning message if this CLI is used against Azure DevOps server

roshan-sy:
Added warning 
**Support for Azure DevOps Server, is in preview and under development**

PR #1160, opened 2021-07-20
Added warning message while using the commands with ADO server

roshan-sy:

Fixes: #1151 

To show the warning message code resolves the organization URL.
the resolved URL is then compared with the know patterns.

https://dev.azure.com/*
*.visualstudio.com

If URL does not matches with know patterns the warning message is shown

### DIFF
--- a/azure-devops/azext_devops/dev/common/services.py
+++ b/azure-devops/azext_devops/dev/common/services.py
@@ -356,7 +356,7 @@ def resolve_instance_project_and_repo(
     if repo_required and repo is None:
         _raise_repo_requird_arg_error()
 
-    if check_organization(organization):
+    if check_organization_in_azure(organization):
         logger.warning("Support for Azure DevOps Server, is under development")
 
     return organization, project, repo
@@ -428,7 +428,7 @@ def get_project_id_from_name(organization, project):
     return project
 
 
-def check_organization(organization):
+def check_organization_in_azure(organization):
     startsWith = organization.startswith("https://dev.azure.com/")
     endsWith = organization.rstrip("/").endswith(".visualstudio.com")
     return not (startsWith or endsWith)
--- a/azure-devops/azext_devops/dev/common/services.py
+++ b/azure-devops/azext_devops/dev/common/services.py
@@ -356,11 +356,10 @@ def resolve_instance_project_and_repo(
     if repo_required and repo is None:
         _raise_repo_requird_arg_error()
 
-    if not (organization.startswith("https://dev.azure.com/") or organization.rstrip("/").endswith(".visualstudio.com")):
-        logger.warning("Support for Azure DevOps Server, is in preview and under development")
+    if check_organization(organization):
+        logger.warning("Support for Azure DevOps Server, is under development")
 
     return organization, project, repo
-
 
 def resolve_instance_and_project(detect, organization, project=None, project_required=True):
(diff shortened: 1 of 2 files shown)


gauravsaralMs (commented):
- will be good to have a better PR title and some more details around the logic in description
- will be good to add test cases
- can you share the experience how does this looks to the customer both
  - against hosted
  - against onprem
- we need review from @aaronhallberg on the messaging
  - I will suggest to not say that the support is in preview as we don't plan to make this support GA

roshan-sy:
image

image

roshan-sy:
> * will be good to have a better PR title and some more details around the logic in description
> * will be good to add test cases
> * can you share the experience how does this looks to the customer both
>   
>   * against hosted
>   * against onprem
> * we need review from @aaronhallberg on the messaging
>   
>   * I will suggest to not say that the support is in preview as we don't plan to make this support GA

Working on test cases and changing the message

roshan-sy:
Done with test cases and message changes

Message: 
`Support for Azure DevOps Server, is under development`

Test cases
```
    def test_check_organization_with_dev_url(self):
        showWarning = check_organization(self._TEST_DEVOPS_ORGANIZATION)
        self.assertEqual(False, showWarning)

    def test_check_organization_with_visual_studio_url(self):
        showWarning = check_organization(self._TEST_VISUAL_STUDIO_ORGANIZATION)
        self.assertEqual(False, showWarning)

    def test_check_organization_with_ado_server_url(self):
        showWarning = check_organization(self._TEST_ADO_SERVER_ORGANIZATION)
        self.assertEqual(True, showWarning)
```

azure-devops/azext_devops/dev/common/services.py
```diff
@@ -423,6 +426,10 @@ def get_project_id_from_name(organization, project):
         return team_project.id
     return project
 
+def check_organization(organization):
```
gauravsaralMs, line 429: check_organization_in_azure

gauravsaralMs: name

roshan-sy: Method name changed

roshan-sy: image

Done

azure-devops/azext_devops/dev/common/services.py
```diff
@@ -355,6 +355,10 @@ def resolve_instance_project_and_repo(
         _raise_team_project_arg_error()
     if repo_required and repo is None:
         _raise_repo_requird_arg_error()
+
+    if not check_organization_in_azure(organization):
+        logger.warning("Support for Azure DevOps Server, is under development")
```
gauravsaralMs, line 360: @aaronhallberg 
need your input on this string

aaronhallberg: I think we should be more explicit here. Something like "_The Azure DevOps Extension for the Azure CLI does not support Azure DevOps Server._"

gauravsaralMs: @roshan-sy : Let's take this change and we will be good to merge this PR

gauravsaralMs (approved):

azure-devops/azext_devops/dev/common/services.py
roshan-sy: Done

Issue #1132, opened 2021-06-03
[Feature Request] az repos ref delete make object-id parameter optional


t3mi:
**Describe the solution you'd like**
Please consider making `--object-id` optional parameter and fetch the value for it automatically in `az repos ref delete` command.

POLIMATISUNILBABU:
I find one page with the method --> az repos ref delete

# delete the reference
            delete_command = 'az repos ref delete --name {} --object-id {} {}'.format(REF_NAME, created_object_id, REPO_NAME)
            deleted_ref = self.cmd(delete_command).get_output_in_json()
            assert deleted_ref is not None
            assert created_ref['updateStatus'] == 'succeeded'
            assert created_ref['success'] is True


--> I commented the existing object id parameter
# assert created_object_id is not None
Created new object id like below
 assert created_object_id is None

gauravsaralMs:
@POLIMATISUNILBABU : please explore the working of [az repos ref list](https://docs.microsoft.com/en-us/cli/azure/repos/ref?view=azure-cli-latest#az_repos_ref_list)

for a given ref name we should be able to detect the object id using the logic present in [az repos ref list](https://docs.microsoft.com/en-us/cli/azure/repos/ref?view=azure-cli-latest#az_repos_ref_list)

PR #1166, opened 2021-08-03
Fixed: Made object id optional

roshan-sy:

Fixes: #1132

Made object id optional while deleting the refs (i.e. heads and tags)

Changes done: 
1. Flow:  Object id is resolved on the go and provided it to API, As Object Id is not optional in API.
2. Code: E2E test case added specifically for delete ref (Create and delete ref)

### DIFF
--- a/azure-devops/azext_devops/dev/repos/ref.py
+++ b/azure-devops/azext_devops/dev/repos/ref.py
@@ -56,7 +56,7 @@ def create_ref(name, object_id, repository=None, organization=None, project=None
                               project=project)[0]
 
 
-def delete_ref(name, object_id, repository=None, organization=None, project=None, detect=None):
+def delete_ref(name, object_id=None, repository=None, organization=None, project=None, detect=None):
     """Delete a reference.
     :param str name: Name of the reference to delete (example: heads/my_branch).
     :param str object_id: Id of the reference to delete.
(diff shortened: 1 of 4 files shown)


azure-devops/azext_devops/dev/repos/ref.py
```diff
@@ -70,6 +70,16 @@ def delete_ref(name, object_id, repository=None, organization=None, project=None
...
         repo=repository)
     client = get_git_client(organization)
+
+    if object_id is None:
+        ref = client.get_refs(repository_id=repository,
+                            project=project,
+                            filter=name)
+        if not ref:
+            logger.error('ref not found')
+            return
```
gauravsaralMs, line 80:
```suggestion
            raise CLIError("Failed to find object_id for ref " + name + ". Please provide object_id.")
```

azure-devops/azext_devops/dev/repos/ref.py
```diff
@@ -70,6 +70,16 @@ def delete_ref(name, object_id, repository=None, organization=None, project=None
         project=project,
         repo=repository)
     client = get_git_client(organization)
+
+    if object_id is None:
+        ref = client.get_refs(repository_id=repository,
+                            project=project,
+                            filter=name)
+        if not ref:
```
gauravsaralMs, line 78:
```suggestion
        if not ref or len(ref) > 1:
```

gauravsaralMs: just to be sure we are not getting multiple id's from service because in that case we can't be sure which one to delete

gauravsaralMs (commented):
I would suggest adding some UTs also, specially for negative cases

roshan-sy:
Working on Adding more Unit test cases

gauravsaralMs:
@roshan-sy : unit tests are present in https://github.com/Azure/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/test/repos/test_ref.py

btw: nice work adding E2E test cases, the reason we need unit test cases as well is that unit test cases are easier/ fast to run and can be used to validate small changes

one unit test is already present for delete_ref

https://github.com/Azure/azure-devops-cli-extension/blob/211d0c58f1ac93b7a645b538e5c4da4b850c0f2e/azure-devops/azext_devops/test/repos/test_ref.py#L87-L96

roshan-sy:
Got it, I'll add the UT as well,

gauravsaralMs (approved):

Issue #1136, opened 2021-06-10
 'GetDefinitionsResponseValue' has no len() on pipelines update


josiahlaivins:
## Describe the bug

**Command Name**
`az pipelines update
Extension Name: azure-devops. Version: 0.18.0.`

**Errors:**
```
The command failed with an unexpected error. Here is the traceback:
object of type 'GetDefinitionsResponseValue' has no len()
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 231, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 657, in execute
    raise ex
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 720, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 712, in _run_job
    return cmd_copy.exception_handler(ex)
  File "/home/svi_core_2/.azure/cliextensions/azure-devops/azext_devops/dev/common/exception_handler.py", line 31, in azure_devops_exception_handler
    reraise(*sys.exc_info())
  File "/opt/az/lib/python3.6/site-packages/six.py", line 703, in reraise
    raise value
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 691, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 328, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/command_operation.py", line 121, in handler
    return op(**command_args)
  File "/home/svi_core_2/.azure/cliextensions/azure-devops/azext_devops/dev/pipelines/pipeline_create.py", line 186, in pipeline_update
    id = get_definition_id_from_name(name, pipeline_client, project)
  File "/home/svi_core_2/.azure/cliextensions/azure-devops/azext_devops/dev/pipelines/build_definition.py", line 106, in get_definition_id_from_name
    if len(definition_references) == 1:
TypeError: object of type 'GetDefinitionsResponseValue' has no len()

```

## To Reproduce:
[...]

PR #1206, opened 2021-09-16
Removed name property from update command

roshan-sy:

 Edit
 Fixes: #1136 

Changes:
1. Removed name property from update command As its deprecated from the API side

### DIFF
--- a/azure-devops/azext_devops/dev/pipelines/pipeline_create.py
+++ b/azure-devops/azext_devops/dev/pipelines/pipeline_create.py
@@ -155,7 +155,7 @@ def pipeline_create(name, description=None, repository=None, branch=None, yml_pa
                               project=project)
 
 
-def pipeline_update(id=None, description=None, new_name=None,  # pylint: disable=redefined-builtin
+def pipeline_update(id, description=None, new_name=None,  # pylint: disable=redefined-builtin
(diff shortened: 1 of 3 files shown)


azure-devops/azext_devops/dev/pipelines/pipeline_create.py
```diff
@@ -155,12 +155,10 @@ def pipeline_create(name, description=None, repository=None, branch=None, yml_pa
                               project=project)
 
 
-def pipeline_update(name=None, id=None, description=None, new_name=None,  # pylint: disable=redefined-builtin
+def pipeline_update(id=None, description=None, new_name=None,  # pylint: disable=redefined-builtin
```
gauravsaralMs, line 158: we should remove `id=None` and only keep `id` 
because now id is mandatory

then this check can also be removed
```python
 if id is None:
        raise CLIError("--id argument must be supplied for this command.")
```

roshan-sy: Got it

gauravsaralMs (approved):
Let's make a note to call out this as a breaking change in next release

@roshan-sy  can you create a issue to track next release and put the note there?

PR #1226, opened 2021-11-08
Fix broken az devops auth

roshan-sy:

### DIFF
--- a/azure-devops/azext_devops/dev/common/services.py
+++ b/azure-devops/azext_devops/dev/common/services.py
@@ -135,7 +135,7 @@ def get_token_from_az_logins(organization, pat_token_present):
         for key, dummy_value in tenantsDict.items():
             try:
                 logger.debug('trying to get token (temp) for tenant %s and user %s ', key[0], key[1])
-                token = get_token_from_az_login(profile, key[1], key[0])
+                token = get_token_from_az_login(profile, key[0])
                 credentials = BasicAuthentication('', token)
 
                 if skipValidateToken is True:
@@ -152,14 +152,17 @@ def get_token_from_az_logins(organization, pat_token_present):
     return ''
 
 
-def get_token_from_az_login(profile, user, tenant):
+def get_token_from_az_login(profile, tenant):
     try:
-        creds, subscription, tenant = profile.get_raw_token(
+        raw = profile.get_raw_token(
             resource='499b84ac-1321-427f-aa17-267ca6975798', tenant=tenant)
+        creds=raw[0]
         auth_token = creds[1]
         return auth_token
     except BaseException as ex:  # pylint: disable=broad-except
         logger.debug('not able to get token from az login')
+        logger.debug(ex, exc_info=True)
+        return ""
 
 
 def _get_connection(organization, credentials):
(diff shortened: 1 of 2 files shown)


gauravsaralMs:
@roshan-sy 
- can you take a look at this file https://github.com/Azure/azure-devops-cli-extension/blob/master/azure-devops/azext_devops/azext_metadata.json (and it's history) and see if we should update that as well ?
- can you share the validation done for this pull request ?

gauravsaralMs (approved):

azure-devops/azext_devops/dev/common/services.py
```diff
@@ -152,9 +152,12 @@ def get_token_from_az_logins(organization, pat_token_present):
     return ''
 
 
-def get_token_from_az_login(profile, user, tenant):
+def get_token_from_az_login(profile, tenant):
```
jiasli, line 155: `get_raw_token` supports **user**, **service principal**, **managed identity** accounts.

By removing `user` from `get_token_from_az_login`, `get_raw_token` will only return token for the currently selected account:

```
1. sub1, user1, from tenant1 - default
2. sub2, user2, from tenant2
```

This will cause `get_raw_token` to get a token for `user1` in `tenant2`.

Consider using `subscription` as the key for `get_raw_token`.

roshan-sy: Tokens are independent of subscriptions. as ADO accounts are not linked to any azure subscriptions.

jiasli: I understand, but Azure CLI uses subscription as the primary key. If you don't specify `subscription` as the key for `get_raw_token`, matched `user` and `tenant` may occur.

jiasli: The old implementation of `get_access_token_for_resource` doesn't work for service principal an managed identity, that's why it is not recommended and removed in the MSAL migration:

https://github.com/Azure/azure-cli/blob/14cc787d0f58bc649d402b486fdecc5625eee9ac/src/azure-cli-core/azure/cli/core/_profile.py#L531-L535

```py
    def get_access_token_for_resource(self, username, tenant, resource):
        tenant = tenant or 'common'
        _, access_token, _ = self._creds_cache.retrieve_token_for_user(
            username, tenant, resource)
        return access_token
```

jiasli: User/tenant isolation was proposed in https://github.com/Azure/azure-cli/issues/15005. After that, it will make sense to make `get_raw_token` support user+tenant.

jiasli:
I don't think it is a good idea to try every Azure CLI account, as this may cause **unexpected user credential to be used**. 

Instead, only the default/selected **(subscription, tenant, user)** combination should be used to get an access token. If that token doesn't have access, we should explicitly switch account with `az account set`.

cveld:
Maybe good next time to have a good PR description. Now it is not clear to me:
- What was the original problem
- What is the proposed solution
- What are the test cases that are run 

In my case I am still not able to use AAD authentication against Azure DevOps. Using az cli 2.30.0 and devops extension 0.22. Also see https://github.com/Azure/azure-cli/issues/20529

jiasli:
> Maybe good next time to have a good PR description.

Completely agree.

Issue #1112, opened 2021-03-22
Support for paging using az devops invoke when continuation token is in header


Tiberriver256:
Example command:

```sh
az devops invoke --area distributedtask --resource jobrequests --route-parameters poolId={myPoolId} --organization {myOrg} --query-parameters `$top=1
```

will translation to the following URL:

```
{myOrg}/_apis/distributedtask/pools/{myPoolId}/jobrequests?$top=1
```

This works as expected and returns a JSON response to the console. The problem is that, if I need to get to the next page, the server sends the continuation token as a header instead of embedded in the JSON body response and az devops invoke doesn't expose response headers at all, that I can see.

roshan-sy:
Looks like API do not support continuation token.

@Tiberriver256 
Can you please check

Tiberriver256:
Hi roshan, sorry for the confusion. Could you use the build definitions list endpoint as an example?

https://docs.microsoft.com/en-us/rest/api/azure/devops/build/definitions/list?view=azure-devops-rest-6.0

There are several endpoints that use the continuation token in the header as a paging mechanism so I can pick another if you like.

roshan-sy:
Fixed.

Will be available after next release.

PR #1243, opened 2022-01-18
Added continuous-token as response

roshan-sy:

Fixes: #1112 

Modified the response to include **continuous-token**

### DIFF
--- a/azure-devops/azext_devops/dev/team/invoke.py
+++ b/azure-devops/azext_devops/dev/team/invoke.py
@@ -105,7 +105,8 @@ def invoke(area=None, resource=None,
                             route_values=route_values,
                             media_type=media_type,
                             accept_media_type=accept_media_type,
-                            content=request_body)
+                            content=request_body,
+                            continuation_token = "||10/31/2018 8:49:39 PM")
     logger.info('content type header')
     logger.info(response.headers.get("content-type"))
     is_content_available = True
--- a/azure-devops/azext_devops/dev/team/invoke.py
+++ b/azure-devops/azext_devops/dev/team/invoke.py
@@ -117,7 +117,7 @@ def invoke(area=None, resource=None,
         response_dict = response.json()
         response_dict["continuation_token"] = response.continuation_token
 
-        return response_dict;
+        return response_dict
 
     # Only handle out file scenario if the content is available (content-type is not None)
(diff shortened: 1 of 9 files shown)


azure-devops/azext_devops/devops_sdk/client.py
```diff
@@ -102,6 +102,7 @@ def _send(self, http_method, location_id, version, route_values=None,
         if Client._session_header_key in Client._session_data and Client._session_header_key not in headers:
             headers[Client._session_header_key] = Client._session_data[Client._session_header_key]
         response = self._send_request(request=request, headers=headers, content=content, media_type=media_type)
+        response.continuation_token = response.headers.get('X-MS-ContinuationToken')
```
aasiddiq, line 105: Maybe it is included someplace else but when we are making a subsequent call don't we need to send the continuation token in request as well?

roshan-sy: Checking

roshan-sy: @aasiddiq 

It is not there actually, the "invoke" command does not have support for continuous token.
We can close this one and I can create a separate "Feature Request" to support "continuous token" in the request header.
And let's track it separately.

roshan-sy:
Canceling pool request as more changes are required

roshan-sy:
Adding sample requests with Changes

image
image

azure-devops/azext_devops/devops_sdk/client.py
roshan-sy: "Clarification"
Invoke supports continuation token but not in the header, It support as a Query string.

azure-devops/azext_devops/dev/team/invoke.py
```diff
@@ -114,7 +114,10 @@ def invoke(area=None, resource=None,
         logger.info('Content type header is None.')
         is_content_available = False
     elif 'json' in response.headers.get("content-type") and not out_file:
-        return response.json()
+        response_dict = response.json()
+        response_dict["continuation_token"] = response.continuation_token
```
aasiddiq, line 118: Since we are adding the token to the response in `client.py` do we still need this?

aasiddiq (approved):

PR #1263, opened 2022-03-10
fix azure pipelines jobs dependson configuration

### LINKED ISSUE #1249: Azure pipelines job dependsOn are not explicit
As a learning experience I have cloned this project to Azure DevOps and found inconsistencies in the Azure Pipelines. 

The pipelines break when only a single build agent is available because the job order is not explicit. From what I have seen mostly "Run_Style_Check" is run before the builds artifacts are created in the jobs 'Build_Publish_Azure_CLI_Test_SDK' and 'Build_Publish_Azure_DevOps_CLI_Extension' fairly consistently. Here is a screenshot to illustrate:
image

For example in [azure-pipelines-merge.yml](https://github.com/Azure/azure-devops-cli-extension/blob/628525c722e946cd97e4212458b548bfe3bdefe6/.azure-pipelines/azure-pipelines-merge.yml):
job: 'Run_Test_Mac_Azure_CLI_Released_Version' should depend on both dependencies but has one.
job: 'Code_Coverage' should depend on both dependencies but has one.
job: 'Run_Style_Check' is missing depends on for both jobs
job: 'Run_Test_From_Old_Release' should depend on both dependencies but has one.

Same for other pipelines.

This dependsOn is required because the templates/download-install-local-azure-test-sdk.yml and templates/download-install-local-azure-devops-cli-extension-with-pip.yml are directly or indirectly used by the jobs. Both jobs download artifacts created by referenced jobs.

You have probably never encountered this because you have multiple build agents available and the reference build tasks complete quickly, before any build artifact downloads are executed. This may still break when the build and\or upload process is delayed or when less build agents are available. Anyways, I think it the job order should be made explicit in the pipeline yaml, so logging this issue for you to consider.

I can send a PR for the dependsOn fixes if needed.

A more DRY solution could be to use stages, for example "build" and "testing" or to use a job template (with it's dependencies included in the template) instead of a steps template. But that might require more changes to the pipelines than strictly needed to fix this.

moerketh:
Fixes #1249 (Azure pipelines job dependsOn are not explicit).

 - [X] : This PR has a corresponding issue open in the Repository.

### DIFF
--- a/.azure-pipelines/azure-pipelines-create-release.yml
+++ b/.azure-pipelines/azure-pipelines-create-release.yml
@@ -39,7 +39,7 @@ jobs:
   - template: templates/build-publish-azure-cli-test-sdk.yml
 
 - job: 'Run_Test'
-  dependsOn : Build_Publish_Azure_CLI_Test_SDK
+  dependsOn : [ 'Build_Publish_Azure_CLI_Test_SDK', 'Build_Publish_Azure_DevOps_CLI_Extension']
   pool:
     vmImage: 'macOS-10.15'
 
--- a/.azure-pipelines/azure-pipelines-merge.yml
+++ b/.azure-pipelines/azure-pipelines-merge.yml
@@ -87,7 +87,7 @@ jobs:
       pythonVersion: '3.x'
 
 - job: 'Run_Test_Mac_Azure_CLI_Released_Version'
-  dependsOn : Build_Publish_Azure_CLI_Test_SDK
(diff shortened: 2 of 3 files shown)


v-soujanya:
@moerketh Hi, we are seeing the same issue with azure-pipelines-pr.yml, below are the errors occurring. By approving this PR "fix azure pipelines jobs dependson configuration #1263 " will be able to resolve this pipeline issue?
image
image

moerketh:
@v-soujanya  Hi! No, I don't think so. Looks like the pipeline needs to be updated to work with a newer python version. Although I would suggest to fix this one too :)

PR #1370, opened 2023-10-20
Auto-populate PR description from commit message

### LINKED ISSUE #695: [Feature Request] Prefill PR description with commit message
**Is your feature request related to a problem? Please describe.**
When I create a PR through the UI for a branch that contains only a single commit relative to the target, the description is set to the commit message. With the CLI Extension (`az repos pr create`) in the same situation the description remains empty.

**Describe the solution you'd like**
When creating a PR through the CLI Extension, ideally I would like the first line of the commit message to be taken as the PR title (already the case), and the third line onward set as the description. I could also live with the behavior in which the complete commit message is taken as the PR description (as is the case in the UI now).

**Additional context**
In my team we have the workflow where we try to squash related commits already on the client. We value well crafted commit messages in line with git recommendations, which provide a concise title (up till 50 chars on the first line) and a more detailed description (separated by a blank line from the title). When a PR is initiated from a pushed commit, it would save time when the PR description is already pre-set from the commit message.

ajryan:
Extends the behavior that auto-populates title when there is a single commit to also copy the entire commit comment to PR description. Resolves #695.

### DIFF
--- a/azure-devops/azext_devops/dev/repos/pull_request.py
+++ b/azure-devops/azext_devops/dev/repos/pull_request.py
@@ -187,20 +187,34 @@ def create_pull_request(project=None, repository=None, source_branch=None, targe
         pr.work_item_refs = resolved_work_items
     pr = client.create_pull_request(git_pull_request_to_create=pr, project=project,
                                     repository_id=repository)
+    # if title or description are not provided and there is a single commit, we will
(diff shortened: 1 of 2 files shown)


v-anvashist:
Hi @ajryan  We are planning to publish next CLI release.

Can you please assist us here to test this PR?

PR #1495, opened 2026-05-15
Align ELM migration enums and skip-validation policies with renamed server contract

itshawi:
## Summary

The ELM service renamed several `MigrationStatus` enum members and added more flags to the
`SkipValidationOptions` flag set. This PR aligns the CLI with the new server contract.

Testing:

I have done some E2E testing and made sure these new changes are working as expected

### DIFF
--- a/azure-devops/azext_devops/dev/migration/migration.py
+++ b/azure-devops/azext_devops/dev/migration/migration.py
@@ -46,8 +46,10 @@ _SKIP_VALIDATION_POLICIES = {
     'maxpullrequestsize': 16,
     'maxpushpacksize': 32,
     'maxreferencenamelength': 64,
-    'maxreposize': 128,
     'targetrepositorydoesnotexist': 256,
+    'sourcerepositorycontainslfsobjects': 512,
+    'sourcerepositorynotreadonly': 1024,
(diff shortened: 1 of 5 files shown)


azure-devops/azext_devops/dev/migration/migration.py
```diff
@@ -58,14 +60,22 @@
...
     'synchronization',
     'cutover'
 }
+
+_MIGRATION_STATUS_VALUES = {
+    'active': 0,
+    'completed': 1,
+    'failed': 2,
+    'paused': 3,
+}
```
itshawi, line 78: We're introducing _MIGRATION_STATUS_VALUES as a name → integer lookup because the ELM service rolls out in rings, so during deployment some rings still run the old enum names (e.g. Suspended) while others run the new ones (e.g. Paused). When we call the API with the integer enum value instead of the name, the request works against every ring uniformly — the integer (3 for paused, 0 for active, etc.) is stable across the rename, while the string name is not.

mishrarish (approved):

itshawi:
> please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.

@microsoft-github-policy-service agree company="Microsoft"

PR #1511, opened 2026-08-06
Fix apiVersionToFloat crashing on preview API versions with a sub-version

### LINKED ISSUE #946: [Bug] az devops invoke doesn't support preview API versions
**Describe the bug**
I was trying to discover feed ids for our projects (since the feed ids I have to use for yaml pipelines aren't easily discoverable) using the CLI, but it doesn't support the preview API version I need: GET https://feeds.dev.azure.com/{organization}/{project}/_apis/packaging/feeds?api-version=5.1-preview.1

**To Reproduce**
Azure Cli Version: 2.0.76
Azure-Devops extension version: 0.17.0

Try to invoke a preview API, for example:
```
az devops invoke --organization "https://feeds.dev.azure.com/{organization}/{project}" --area packaging --resource feeds --api-version "5.1-preview.1"
```
View the output:
```
could not convert string to float: '5.1.1'
```

**Additional context**
The invoke command appears to be trying to convert the `api-version` into a float, even though it should just be left as a string.

Also, the invoke command also appears to be missing several URL parameters from the [documented](https://docs.microsoft.com/en-us/rest/api/azure/devops/?view=azure-devops-rest-5.1) form: `VERB https://{instance}[/{team-project}]/_apis[/{area}]/{resource}?api-version={version}`. For example, `team-project` is necessary now for packaging, and I'm not sure if `feeds.` should be part of the `{instance}` or not. It would also be nice if I could just provide a full API url and have `az devops invoke` provide authentication for that url.

### LINKED ISSUE #1454: Allow az devops invoke to support preview API versions
Allow az devops inovke to support preview API version.

At present:

`az devops invoke --organization "https://feeds.dev.azure.com/{organization}/{project}" --area packaging --resource feeds --api-version "5.1-preview.1"
`
return the following output:
`could not convert string to float: '5.1.1'`

This issue have been report on ticket https://github.com/Azure/azure-devops-cli-extension/issues/946

allan-oliveira:
## Summary

`az devops invoke --api-version <X.Y-preview.N>` fails with `could not convert string to float: '<X.Y.N>'`.

`apiVersionToFloat` only strips the literal `-preview` substring:

```python
def apiVersionToFloat(apiVersion):
    apiVersion = apiVersion.replace('-preview', '')
    return float(apiVersion)
```

For an input like `7.1-preview.1`, this leaves `7.1.1` behind (the sub-version suffix stays attached), which `float()` can't parse.

## Fix

Split on `-preview` and keep only the part before it, so the whole preview marker (including any sub-version) is dropped:

```python
def apiVersionToFloat(apiVersion):
    apiVersion = apiVersion.split('-preview')[0]
    return float(apiVersion)
```

This is the only call site of `apiVersionToFloat` in the codebase, and the parsed float is only used for a local version comparison against `resource_location.max_version` (line 91) — the original, unmodified `api_version` string is still passed through unchanged to the actual HTTP client call (line 103), so this only fixes the local comparison logic without touching wire behavior.

## Test plan

- Added `azext_devops/tests/latest/team/test_invoke.py` covering plain versions, `X.Y-preview`, `X.Y-preview.N`, and multi-digit sub-versions.
- Manually verified the fixed function against the exact failing input from the linked issues.

Fixes #946
Fixes #1454

### DIFF
--- a/azure-devops/azext_devops/dev/team/invoke.py
+++ b/azure-devops/azext_devops/dev/team/invoke.py
@@ -135,7 +135,7 @@ def invoke(area=None, resource=None,
 
 
 def apiVersionToFloat(apiVersion):
-    apiVersion = apiVersion.replace('-preview', '')
+    apiVersion = apiVersion.split('-preview')[0]
 
     return float(apiVersion)
 
--- a/azure-devops/azext_devops/tests/latest/team/test_invoke.py
+++ b/azure-devops/azext_devops/tests/latest/team/test_invoke.py
@@ -1,0 +1,24 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import unittest
+
+from azext_devops.dev.team.invoke import apiVersionToFloat
+
+
+class TestApiVersionToFloat(unittest.TestCase):
+
+    def test_plain_version(self):
+        self.assertEqual(apiVersionToFloat('7.1'), 7.1)
+
+    def test_preview_version_without_subversion(self):
+        self.assertEqual(apiVersionToFloat('7.1-preview'), 7.1)
+
+    def test_preview_version_with_subversion(self):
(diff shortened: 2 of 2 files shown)


allan-oliveira:
@microsoft-github-policy-service agree

PR #1512, opened 2026-08-10
Remove legacy GitHub auth args from ELM migrations create (device flow only)

bhuvanshahMSFT:
Makes interactive GitHub device flow the only auth path for `az devops migrations create`.

Changes:
- Remove `--service-endpoint-id` and `--github-token` / `ELM_GITHUB_TOKEN` args. The server now supports third-party app auth, so the CLI resolves the GitHub user token via interactive device flow only.
- Keep `--pipeline-service-connection-id` (unchanged).

Testing:
- 102 unit tests pass in test_migration.py.
- Validated end-to-end against production (validate-only migration mseng/1ES → msft.ghe.com/1ES); server resolved the target owner from the device-flow token, no service connection required.

### DIFF
--- a/azure-devops/azext_devops/dev/migration/_help.py
+++ b/azure-devops/azext_devops/dev/migration/_help.py
@@ -38,7 +38,7 @@ def load_migration_help():
     helps['devops migrations create'] = """
     type: command
     short-summary: Create a migration for a repository.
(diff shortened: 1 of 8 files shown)


bhaveshbhati (approved):
LGTM

